PR #22, opened 2017-04-30
Mobile home revisions

dannypernik:


openCurrents/templates/home.html
```diff
@@ -250,6 +249,20 @@ <h4 class="white half-margin-bottom no-margin-top">Who benefits?</h4>
...
+          <a href="#" class="accordion-title">
+            <div class="center">
+              <p class="white half-margin-bottom lead">Everyone.</p>
+              <button id="all-button" class="benefit-icon button clear" autofocus>
+                <i class="fa fa-2x fa-globe"></i>
+              </button>
+            </div>
+          </a>
+          <div class="accordion-content" data-tab-content>
+            When good work is rewarded, more good is given. Society as a whole benefits.
```
nickolashe: Can we change "society" to "community"? I think society sounds a bit less specific.

openCurrents/templates/home.html
```diff
@@ -20,7 +20,7 @@
       <div class="row center">
         <div class="input-middle small-12 medium-11 small-centered columns">
           <h5 class="white one-margin-top half-margin-bottom">Introducing the world's first digital currency generated by social action.</h5>
```
nickolashe: @dannypernik Have we actually verified that this statement is true?

nickolashe:
@dannypernik Just the two comments above. I assume you've tested all the UI changes

dannypernik:
Good question. No we haven't. When I tried googling "currency created by
social good" (maybe with a typo) I got this:

openCurrents/templates/home.html
nickolashe: Bump

dannypernik: @nickolashe I know Max prefers society to community. I like using community most of the time because it can refer to local communities or the global community. However in this context when we're talking about the world as a whole, I think it makes sense to use the general term society.

dannypernik:
@nickolashe note the additional commit. Please test that the back/forward buttons are working appropriately for you. User is actually able to progress by hitting forward from initial page load, which is a little weird, but I'm just relieved I got the buttons working.

openCurrents/templates/home.html
nickolashe: @dannypernik Ok, so based on our conversation are we making the emphasis on the uniqueness of the digital aspect then?

openCurrents/templates/home.html
nickolashe: We can equally say "community as a whole" and sound more concrete.

nickolashe:
@dannypernik Ok, will get this new commit tested.

nickolashe:
@dannypernik Back button is working for me (on the "Tell me more" flow). Do you plan to address my other comments before we merge this?

dannypernik:
@nickolashe can you take a look and see if home page flow is looking good? If so, can we merge this? The only remaining question from this pull request, I believe, is whether to say "community as a whole" or "society as a whole". I prefer "society" but we can discuss it. Regardless, I don't think it should hold up a merge.

PR #29, opened 2017-05-21
Hour tracking

nickolashe:


openCurrents/models.py
```diff
@@ -72,3 +48,71 @@ def __unicode__(self):
...
+
+    # created / updated timestamps
+    date_created = models.DateTimeField('date created', auto_now_add=True)
+    date_updated = models.DateTimeField('date last updated', auto_now=True)
+
+    def __unicode__(self):
+        return ' '.join([self.user.username, '\'s account'])
+
+
+class Project(models.Model):
```
nickolashe: @dannypernik Can you take a look at this? I propose we have a concept of a project and another of a project event (which is a specific occurrence with a location and time). Admin would be able to create both. And when creating an event, should be able to pick an existing project.

dannypernik: @nickolashe ok so a project would basically be an action in the form of "Let's (do some good)" and events would be specific time and date? I'm not sure that's going to make sense to users... Is the goal here to make it easy to create new similar events? If so, saving templates seems easiest, and when an admin wants to reuse a template, they only need to update the date and time. Is there another reason for projects + events?

nickolashe: @dannypernik This would not have any user-facing consequences. It is possible to do it both ways, but I think this way is a lot cleaner. Since you had templates that listed projects both in a event-agnostic and event-aware way, this would simplify things a lot in that respect.

dannypernik:
@nickolashe ok if this enables us to show multiple locations for the same
project as either a single event or multiple events, then this makes sense

openCurrents/templates/create-project.html
```diff
@@ -11,10 +11,10 @@
     <link href="{% static 'css/jquery-ui-datepicker.css' %}" rel="stylesheet" type="text/css">
```
nickolashe: @dannypernik Lets use this template to:
1. Pick up the new form syntax
2. Agree on the project / event distinction

My vote is for the admin to create a project once, then use it to create events for each date.
A project would be tied to a location.

dannypernik:
@nickolashe a project would be tied to a location? So creating a different
project for each location? Perhaps I still don't see this the way you are.
I think having both projects and events for admins to create will be
confusing.

nickolashe:
@dannypernik Lets talk about this live

openCurrents/forms.py
```diff
@@ -52,9 +52,24 @@ def to_python(self, value):
 
 
 class UserSignupForm(forms.Form):
-    user_firstname = forms.CharField()
-    user_lastname = forms.CharField()
-    user_email = forms.EmailField()
+    user_firstname = forms.CharField(
```
nickolashe: @dannypernik If you are still having a problem getting text input to work, here is a working example.

openCurrents/templates/live-dashboard.html
```diff
@@ -137,34 +158,73 @@ <h4 class="popup-title">Select your location</h4>
...
+                url: "{% url 'openCurrents:event_register_live' eventid=event.id %}",
+                type: 'POST',
+                data : {
+                  csrfmiddlewaretoken: document.getElementsByName('csrfmiddlewaretoken')[0].value,
+                  userid: userid
+                },
+                dataType : "json",
+              }).complete(function( data ) {
+                console.log('success');
+                $('#registered-volunteers').append(`
```
nickolashe: @dannypernik As I noted elsewhere before, please use the following formatting when quoting html inside javascript.

PR #100, opened 2017-06-09
live-dashboard: Unit tests

karbmk:
Test case for event checkin redirection to live-dashboard added.


openCurrents/tests.py
```diff
@@ -1,3 +1,94 @@
...
+            'username': 'testuser',
+            'password': 'secret'}
+        self.user = User.objects.create_user(**self.credentials)
+
+    def test_event_checkin(self):
+        """TestCase for successful checkin"""
+        # send login data
+        self.client.login(username='testuser', password='secret')
+        form_data = { 'userid':2,'checkin':True }
+        request = self.client.post('/event_checkin/1/', form_data, follow=True)
```
nickolashe: @karbmk What's the meaning of the `follow` param, I am curious?

openCurrents/tests.py
```diff
@@ -1,3 +1,94 @@
...
+        #request.user = self.user
+        #pk = '1'
+        #response = event_checkin(request,pk=pk)
+        #self.assertEqual(response.status_code, 200)
```
nickolashe: @karbmk Any particular reason for commenting this out? Did this not pass?

openCurrents/tests.py
```diff
@@ -1,3 +1,94 @@
...
+    """TestSuite for event checkin"""
+    def setUp(self):
+        """Setting up the user to test event checkin"""
+        self.factory = RequestFactory()
+        self.credentials = {
+            'username': 'testuser',
+            'password': 'secret'}
+        self.user = User.objects.create_user(**self.credentials)
+
+    def test_event_checkin(self):
```
nickolashe: @karbmk I think we should also assert the corresponding record in UserTimeLog table was created

nickolashe:
@karthik making sure you are seeing my comments on the PR

openCurrents/tests.py
karbmk: Hi, sorry for the late reply. I will take a look on it for sure

openCurrents/tests.py
karbmk: follow parameter is basically to follow the assertions after we hit the url like is the status code 401 or is the page returning a message etc..

openCurrents/tests.py
karbmk: It did fail because the assertEqual did'nt work. I will take a look on that

karthik:
@nickolashe I am not the person you are looking for.

PR #108, opened 2017-06-10
admin-profile: list past events

dannypernik:


nickolashe:
@dannypernik Merged my changes to provide events_past. Note that we also need a button next to each event linking to live dashboard. You can use event.id to get the id.

dannypernik:
@nickolashe made the change, LGTM

PR #130, opened 2017-07-05
Manual Hour Tracking

karbmk:
Manual tracking


nickolashe:
@karbmk can we change the title of the PR to reflect the feature being developed pls?

openCurrents/forms.py
```diff
@@ -231,6 +231,72 @@ class EventRegisterForm(forms.Form):
     )
 
 
+class TrackVolunteerHours(forms.Form):
```
nickolashe: looks good

openCurrents/models.py
```diff
@@ -99,6 +99,43 @@ def __unicode__(self):
         ])
 
 
+class ManualTracking(models.Model):
```
nickolashe: @karbmk we may not need this new model, you should be able to use UserTimeLog model and set is_verified to false in this case

openCurrents/views.py
```diff
@@ -184,9 +192,85 @@ class UserHomeView(LoginRequiredMixin, SessionContextView, TemplateView):
 class VerifyIdentityView(TemplateView):
     template_name = 'verify-identity.html'
 
+class TimeTrackerView(LoginRequiredMixin, SessionContextView, TemplateView):
+    template_name = 'time-tracker.html'
+    form_class = TrackVolunteerHours 
+    success_url = '/time-tracked/'
+    
+    def track_hours(self, form_data):
```
nickolashe: @karbmk can you point me to where you use this function pls?

openCurrents/views.py
```diff
@@ -184,9 +192,85 @@ class UserHomeView(LoginRequiredMixin, SessionContextView, TemplateView):
...
+            time_start = form_data['datetime_start'],
+            time_end = form_data['datetime_end']
+        )
+        track.save()
+
+    def form_valid(self, form):
+        # This method is called when valid form data has been POSTed.
+        # It should return an HttpResponse.
+        data = form.cleaned_data
+        if data['project_name'] in self.project_names:
```
nickolashe: @karbmk i think we can create one event per org for manual tracking and write all hours associated with this event

openCurrents/templates/time-tracker.html
```diff
@@ -0,0 +1,247 @@
...
+
+        <div id="choose-org" class="half-margin-top row center">
+          <div class="center medium-6 small-centered columns">
+
+              <h6 class="subtitle">for</h6>
+
+              <select id="user-orgs" class="two-margin-bottom">
+                <option disabled selected hidden>Select organization</option>
+                <!-- all requested orgs should appear here unless org declines a volunteer -->
+                <option>Bike Austin</option>
```
nickolashe: @karbmk this list is still hardcoded at this point, right?

karbmk: Yes it is

openCurrents/models.py
karbmk: I will work on it and let you know if there are any issues. Thanks

openCurrents/views.py
karbmk: I was calling it in form_valid function below where I have it commented but I ran the code without commenting still din't work.

openCurrents/views.py
karbmk: I think i got it but I'm not sure if what i'm thinking is right. I will get back to you on this.

openCurrents/models.py
nickolashe: @karbmk ok, did my comment make sense?

openCurrents/views.py
nickolashe: @karbmk basically i meant we need some event to associate the hours to if we use the UserTimeLog model for tracking

nickolashe:
@karbmk have your most recent changes been commited to this PR?

nickolashe:
@karbmk Can we close this one now?

karbmk:
Sure @nickolashe since we are working on the other branch we could close this!

PR #131, opened 2017-07-06
Opt in for monthly_updates

tsiology:
Adding monthly_update template checkbox and boolean support.


nickolashe:
@tsiology can we change the title of the PR to reflect the feature being developed pls?

README.md
```diff
@@ -1,4 +1,5 @@
 #Install instructions for the Mac
+#tgodwin
```
nickolashe: ?

openCurrents/views.py
```diff
@@ -804,6 +804,12 @@ def process_email_confirmation(request, user_email):
 
         # create user account
         user_account = Account(user=user, pending=1)
+
+        if form.cleaned_data['monthly_updates']:
```
nickolashe: looks good

nickolashe: @dannypernik @tsiology have we thought about how the users are going to be able to opt out if deploy this?

README.md
dannypernik: We were just testing a git push. Can be removed

openCurrents/views.py
dannypernik: @nickolashe We would have the Unsubscribe button at the bottom of any Mailchimp email. So somehow we have to get the opt-in list to Mailchimp?

README.md
nickolashe: got it

openCurrents/views.py
nickolashe: @dannypernik is there a simple way we can set up a simple opt in/out in our system? may be in profile?

nickolashe: this would be preferred at this point

PR #136, opened 2017-07-06
Check-email page with email displayed after user registers

tsiology:


nickolashe:
@tsiology is this working now?

tsiology:
Yes, with these updates new users are redirected to check-email page which reports the email we sent to

tsiology:
Still work to be done on the resend email button on the check-email page

tsiology:
@nickolashe I'm close to having the Resend button implemented. Before committing that, can we merge this current small PR which just finalizes the check-email redirection?

nickolashe:
@tsiology sure, please go ahead and merge it

PR #139, opened 2017-07-07
Manual tracking

karbmk:
Added the manual tracking of hours for volunteers

### DIFF
--- a/openCurrents/forms.py
+++ b/openCurrents/forms.py
@@ -84,7 +84,6 @@ class EmailVerificationForm(forms.Form):
     user_password = forms.CharField(min_length=8)
     user_password_confirm = forms.CharField(min_length=8)
     verification_token = forms.UUIDField()
-    monthly_updates = forms.BooleanField(initial=False,required=False)
 
     def clean(self):
         cleaned_data = super(EmailVerificationForm, self).clean()
@@ -230,6 +229,69 @@ class EventRegisterForm(forms.Form):
         }),
         max_length=16384
     )
+
+
+class TrackVolunteerHours(forms.Form):
+
+
+    description = forms.CharField(
+        widget=forms.TextInput(attrs={
+            #'rows': '1',
+            'class': 'center large-text',
+            'placeholder': 'Description of work'
+        })
+    )
+    date_start = forms.CharField(
+        #label='on',
+        label = 'Date',
+        widget=forms.TextInput(attrs={
+            'id': 'volunteer-date',
+            'name':'volunteer-date',
+            'placeholder': 'yyyy-mm-dd'
+        })
+    )
+    time_start = forms.CharField(
+        #label='from',
+        label = 'Start time',
+        widget=forms.TextInput(attrs={
+            'id': 'start-time',
+            'name':'',
+            'value': '12:00:00'
+        })
+    )
+    time_end = forms.CharField(
+        #label='to',
+        label = 'End time',
+        widget=forms.TextInput(attrs={
+            'id': 'end-time',
+            'name':'',
+            'value': '12:00:00'
+        })
+    )
+
+    def clean(self):
+        cleaned_data = super(TrackVolunteerHours, self).clean()
+        date_start = cleaned_data['date_start']
+        time_start = cleaned_data['time_start']
+        time_end = cleaned_data['time_end']
+        tz = "America/Chicago"#self.org.timezone
+
+        datetime_start = datetime.strptime(
+            ' '.join([date_start, time_start]),
+            '%Y-%m-%d %I:%M%p'
+        )
+        cleaned_data['datetime_start'] = pytz.timezone(tz).localize(datetime_start)
+
+        datetime_end = datetime.strptime(
+            ' '.join([date_start, time_end]),
+            '%Y-%m-%d %I:%M%p'
+        )
+        cleaned_data['datetime_end'] = pytz.timezone(tz).localize(datetime_end)
+
+        if cleaned_data['datetime_start'] > cleaned_data['datetime_end']:
+            raise ValidationError(_('Start time needs to occur before End time'))
+
+        return cleaned_data
 
 
 class EventCheckinForm(forms.Form):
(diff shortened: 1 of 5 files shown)


openCurrents/forms.py
```diff
@@ -84,7 +84,6 @@ class EmailVerificationForm(forms.Form):
     user_password = forms.CharField(min_length=8)
     user_password_confirm = forms.CharField(min_length=8)
     verification_token = forms.UUIDField()
-    monthly_updates = forms.BooleanField(initial=False,required=False)
```
nickolashe: @karbmk it looks like we'd be reverting Tim's changes with this. have you merged develop?

openCurrents/forms.py
```diff
@@ -232,6 +231,69 @@ class EventRegisterForm(forms.Form):
...
+            'name':'volunteer-date',
+            'placeholder': 'yyyy-mm-dd'
+        })
+    )
+    time_start = forms.CharField(
+        #label='from',
+        label = 'Start time',
+        widget=forms.TextInput(attrs={
+            'id': 'start-time',
+            'name':'',
```
nickolashe: any particular reason for specifying an empty name field here?

openCurrents/forms.py
```diff
@@ -232,6 +231,69 @@ class EventRegisterForm(forms.Form):
...
+            'value': '12:00:00'
+        })
+    )
+
+    def clean(self):
+        cleaned_data = super(TrackVolunteerHours, self).clean()
+        date_start = cleaned_data['date_start']
+        time_start = cleaned_data['time_start']
+        time_end = cleaned_data['time_end']
+        tz = "America/Chicago"#self.org.timezone
```
nickolashe: did the commented line not work? we'd need org timezone here

openCurrents/templates/time-tracker.html
```diff
@@ -10,33 +10,28 @@
...
+        <form action ="" method="post">{% csrf_token %}
         <div id="choose-org" class="half-margin-top row center">
           <div class="center medium-6 small-centered columns">
 
               <h6 class="subtitle">for</h6>
-              
+
               <select id="user-orgs" class="two-margin-bottom">
                 <option disabled selected hidden>Select organization</option>
-                <!-- all requested orgs should appear here unless org declines a volunteer -->
```
nickolashe: looks like this still needs the org list populating the select options here?

openCurrents/templates/time-tracker.html
```diff
@@ -126,37 +133,96 @@ <h3 class="popup-title">Volunteer request</h3>
...
-        
+
         $('#add-org-popup').popup({
           vertical: 'top',
           transition: 'all 0.75s',
           scrolllock: true,
         });
+
+
+          $.ajax
```
nickolashe: any particular reason we are going with ajax request here instead of passing the list to the template?

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
...
+
+        # obtain orgid from the session context (provided by SessionContextView)
+        orgid = context['orgid']
+        self.orgid = orgid
+
+        # context::project_names
+        projects = Project.objects.filter(
+            org__id=orgid
+        )
+        project_names = [
```
nickolashe: why do we need to pass in project names?

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
...
+            a.append(Org.objects.get(id=k).name)
+            k += 1
+        except:
+            i=0
+    print(str(a))
+    return HttpResponse(str(a)[1:-1].replace("'",""),content_type="application/type")
+
+
+@csrf_exempt
+def manual_track_org(request):
```
nickolashe: @karbmk not sure i understand the purpose of this endpoint

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
...
+            for project in projects
+        ]
+
+        context['project_names'] = mark_safe(json.dumps(project_names))
+        self.project_names = project_names
+        logger.info(project_names)
+
+        return context
+
+    def get_form_kwargs(self):
```
nickolashe: this is where we could pass in the org <=> timezone map so the hours use proper timezone

nickolashe:
@karbmk done with my round of comments

openCurrents/forms.py
karbmk: Oh I think I had the forms overwritten. I will revert that. sorry about that

openCurrents/forms.py
karbmk: I actually replicated the field attributes from the html tags which were present in time-tracker.html

openCurrents/forms.py
karbmk: Yep the self.org.timezone isn't working since it gave an error in this part "orgid = kwargs.pop('orgid')" which was there previously.

openCurrents/templates/time-tracker.html
karbmk: Nope this is being handled by the document.ready function if you see in the script below.

openCurrents/forms.py
nickolashe: Yep please see my comment https://github.com/opencurrents/opencurrents/pull/139#discussion_r126257442 on how we can make that happen (`get_form_kwargs` is the function that populates kwargs in forms.py)

openCurrents/forms.py
nickolashe: Is time-tracker missing the name field then?

openCurrents/templates/time-tracker.html
karbmk: Yep I tried passing the template. I was successful in creating the form and UI for the same but I could'nt get the selected option from the UI. It kept on throwing me back to the time-tracker page and error while trying to get the option

openCurrents/views.py
karbmk: This was added because I could'nt get the selected option of the org from the forms and I used this function to pass the org selected by the user to the TimeTrackerView in views

openCurrents/views.py
karbmk: Do you mean something like "def get_form_kwargs(self,orgid):" ?

openCurrents/views.py
karbmk: There is no need for it actually. I thought it needed because we were getting the project above in the track_hours

openCurrents/views.py
nickolashe: Please see an example here:
https://github.com/opencurrents/opencurrents/blob/master/openCurrents/views.py#L371

nickolashe:
@karbmk feel free to close this pr

PR #141, opened 2017-07-08
Manual tracking revisited

karbmk:
Please take a look at it.

### DIFF
--- a/openCurrents/views.py
+++ b/openCurrents/views.py
@@ -9,6 +9,8 @@ from django.db import IntegrityError
 from django.http import HttpResponse
 from django.utils.safestring import mark_safe
 from django.db.models import F, Max
+from django.views.decorators.csrf import csrf_exempt,csrf_protect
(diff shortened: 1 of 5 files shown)


openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
     template_name = 'verify-identity.html'
 
 
-class TimeTrackerView(TemplateView):
+def orgs_list(request):
+    """Lists all the organisations"""
+    i=1
+    a=[]
+    k=1
+    while i:
```
nickolashe: @karbmk why don't we just use Org.objects.all() to obtain the list of orgs?

nickolashe: we are making a separate db call for each org, which is pretty inefficient. also, if the id's end up non-consecutive i think this code will have a problem

nickolashe:
@karbmk feel free to close this pr

PR #149, opened 2017-07-10
process_resend functionality

tsiology:
Allows new users to resend verification email from the check-email page


tsiology:
@nickolashe @dannypernik This PR from Tim to the develop branch adds the Resend email functionality from the check-email page without using a hidden input. 

Need to add explicit ordering on the token list before sending the email. Need to clean out the unnecessary migrations files.

openCurrents/views.py
```diff
@@ -643,6 +644,47 @@ def event_register_live(request, eventid):
 
     return HttpResponse({'userid': userid, 'eventid': eventid}, status=201)
 
+# resend the verification email to a user who hits the Resend button on their check-email page
+def process_resend(request, user_email):
+    form = UserResendForm(request.POST)
```
nickolashe: @tsiology What is the reason we are still using a form here?

openCurrents/views.py
```diff
@@ -643,6 +644,47 @@ def event_register_live(request, eventid):
...
+
+    user = User.objects.get(email=user_email)
+    token_records = Token.objects.filter(email=user_email)
```
nickolashe: @tsiology Another way to go about this is to use django queryset `last` clause and specify `get_latest_by` on the Token model. E.g., see how this is done on UserTimeLog model:
https://github.com/opencurrents/opencurrents/blob/master/openCurrents/models.py#L178

openCurrents/views.py
```diff
@@ -643,6 +644,47 @@ def event_register_live(request, eventid):
...
+                },
+                {
+                    'name': 'TOKEN',
+                    'content': str(token)
+                }
+            ],
+            user_email
+        )
+    except Exception as e:
+        logger.error(
```
nickolashe: In case of error, do we still want to redirect to `check-email`?

dannypernik: @nickolashe I'd say yes. I believe status message is working there.

openCurrents/views.py
tsiology: Thanks, I added get_latest_by to the Token model and implemented .last() in view

https://github.com/opencurrents/opencurrents/blob/tim/openCurrents/views.py#L653

openCurrents/views.py
nickolashe: `check-email` template says the email message was sent, however

openCurrents/views.py
nickolashe: great

openCurrents/views.py
dannypernik: @nickolashe not following here

nickolashe: the template we are redirecting to says the email was sent while we are here because there was an error sending it

PR #151, opened 2017-07-11
merge master

nickolashe:


openCurrents/views.py
```diff
@@ -414,7 +414,7 @@ class EditProjectView(TemplateView):
 
 
 # TODO: prioritize view by projects which user was invited to
-class UpcomingProjectsView(LoginRequiredMixin, SessionContextView, ListView):
+class UpcomingEventsView(LoginRequiredMixin, SessionContextView, ListView):
```
nickolashe: @dannypernik this did it for me as far as renaming the view

dannypernik: @nickolashe  hm ok I tried changing this and urls.py of the same name and it threw me an error when trying to load that template. But if it's working for you then great. Maybe I even misspelled it.

nickolashe: not sure, i thought perhaps you forgot to change one of those

Issue #30, opened 2017-05-29
Implement Django's custom input widgets


nickolashe:
https://docs.djangoproject.com/en/1.11/topics/forms/media/

We can identify common patterns in the UI and define custom widgets, so we don't have to explicitly specify css styling in the template any more.

dannypernik:
@nickolashe I prefer to use css files unless there's some performance advantage?

nickolashe:
that's fine by me

PR #154, opened 2017-07-11
Admin profile event filter

karbmk:
Adding filters to admin-profile and upcoming-event page


openCurrents/forms.py
```diff
@@ -232,6 +231,72 @@ class EventRegisterForm(forms.Form):
     )
 
 
+class TrackVolunteerHours(forms.Form):
+
+
+    CHOICES = (("Select Organisation","Select Organisation"),)
```
nickolashe: @karbmk what is the reason why you chose to populate the orgs through an ajax request to an endpoint rather than directly through django context?

openCurrents/models.py
```diff
@@ -99,6 +99,43 @@ def __unicode__(self):
             self.org.name
         ])
 
+'''
+class ManualTracking(models.Model):
```
nickolashe: can we remove this now?

openCurrents/templates/admin-profile.html
```diff
@@ -59,6 +59,7 @@ <h6 class="inline"> {{ issued_total }}</h6>
               <h6 class="one-margin-bottom">Events happening now</h6>
 
               {% for event in events_current %}
+              {% if event != "ManualTracking" %}
```
nickolashe: @karbmk should not this be `event.project.name`?

nickolashe: also i think we'd need to indent this relative to the `for` clause for consistency

openCurrents/templates/time-tracker.html
```diff
@@ -10,33 +10,29 @@
...
     {% include "header.html" %}
-    
+
     <div id="container" class="js-close-any" canvas="container">
       <div id="wrapper">
         <div class="row">
           <h3 class="title-sub">Track volunteer time</h3>
         </div>
-        
+        <form action ="" method="post">{% csrf_token %}
```
nickolashe: indent `{% csrf_token %}`

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
     template_name = 'verify-identity.html'
 
 
-class TimeTrackerView(TemplateView):
+def orgs_list(request):
+    """Lists all the organisations"""
+    i=1
+    a=[]
+    k=1
+    while i:
```
nickolashe: @karbmk same question as before, i think we should issue a single db call for obtain all orgs through `.all` queryset clause

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
...
+            k += 1
+        except:
+            i=0
+    print(str(a))
+    return HttpResponse(str(a)[1:-1].replace("'",""),content_type="application/type")
+
+
+@csrf_exempt
+def manual_track_org(request):
+    """manually track org"""
```
nickolashe: @karbmk can you explain what this point is doing for me please?

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
...
+        data = form.cleaned_data
+        self.track_hours(data)
+        return redirect('openCurrents:time-tracked')
+
+    
+    def get_context_data(self, **kwargs):
+        context = super(TimeTrackerView, self).get_context_data(**kwargs)
+
+        # obtain orgid from the session context (provided by SessionContextView)
+        orgid = context['orgid']
```
nickolashe: @karbmk where does `orgid` come from?

openCurrents/templates/admin-profile.html
karbmk: I worked with that but I will take a look at it again and also I will make the indentation relative to the for block.

openCurrents/templates/time-tracker.html
karbmk: Sure. This will be done

openCurrents/views.py
karbmk: Sorry. this is the old code. I pulled the latest from develop and it has merged the the old code. I will take care of that.

openCurrents/views.py
karbmk: This is getting the organisation from the user selected option. This is because the html tagged data is not able to get through forms.

nickolashe: got it, but what does it do with the org? i don't see where it is getting persisted.

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
...
+    return HttpResponse(str(a)[1:-1].replace("'",""),content_type="application/type")
+
+
+@csrf_exempt
+def manual_track_org(request):
+    """manually track org"""
+    c = {}
+    c.update(csrf(request));
+    data = request.POST["org_name"]
+    global g_org_name
```
nickolashe: @karbmk lets try to avoid using global variables; things can get very confusing if we do

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
...
+    g_org_name = data
+    return HttpResponse("success",content_type="application/type")
+
+
+class TimeTrackerView(LoginRequiredMixin, SessionContextView, FormView):
     template_name = 'time-tracker.html'
+    form_class = TrackVolunteerHours 
+    success_url = '/time-tracked/'
+    
+    def track_hours(self, form_data):
```
nickolashe: this looks good based on what we discussed

openCurrents/views.py
```diff
@@ -196,8 +199,111 @@ class VerifyIdentityView(TemplateView):
...
+
+    
+    def get_context_data(self, **kwargs):
+        context = super(TimeTrackerView, self).get_context_data(**kwargs)
+
+        # obtain orgid from the session context (provided by SessionContextView)
+        orgid = context['orgid']
+        self.orgid = orgid
+
+        # context::project_names
```
nickolashe: @karbmk the plan was to remove `project_names` altogether, correct?

openCurrents/views.py
karbmk: Yep. If we are able to get the user selected org from the form then we can eliminate that.

openCurrents/views.py
karbmk: Yes.

openCurrents/templates/time-tracker.html
```diff
@@ -188,7 +188,7 @@ <h3 class="popup-title">Volunteer request</h3>
       });
     </script>
 
-    <!-- <script type="text/javascript">
+    <script type="text/javascript">
```
nickolashe: @dannypernik @karbmk what do we need this script for?

openCurrents/forms.py
```diff
@@ -243,11 +243,16 @@ class TrackVolunteerHours(forms.Form):
 
     choices_init = [("select_org","Select organization")]
     choices = [
-        (org.name, org.name)
+        (org.id, org.name)
```
nickolashe: @karbmk lets use org id as the identifier since all db lookups are going to be faster by id (it is an indexed field on the table)

openCurrents/forms.py
```diff
@@ -243,11 +243,16 @@ class TrackVolunteerHours(forms.Form):
...
         for org in Org.objects.all().order_by('name')
     ]
-    choices = choices_init+choices
-    orgs = forms.ChoiceField(choices=choices)
+    choices = choices_init + choices
+    org = forms.ChoiceField(
+        choices=choices,
+        widget=forms.Select(attrs={
+            'id': 'id_org_choice'
```
nickolashe: @karbmk @dannypernik lets specify element id explicitly so as not to rely on the default django choice (since it may change in later versions)

openCurrents/templates/time-tracker.html
```diff
@@ -263,7 +263,7 @@ <h3 class="popup-title">Volunteer request</h3>
       var y = n.getFullYear();
       var m = n.getMonth();
       var d = n.getDate();
-      document.getElementById("start-date").innerHTML = d + " " + months[m] + " " + y;
+      document.getElementById("volunteer-date").innerHTML = d + " " + months[m] + " " + y;
```
nickolashe: @karbmk this was throwing a js error (please check js console for any errors before committing changes)

openCurrents/views.py
```diff
@@ -207,34 +207,35 @@ class TimeTrackerView(LoginRequiredMixin, SessionContextView, FormView):
...
 
         try:
             self.project = Project.objects.get(
-                    org__id=org,
+                    org__id=org.id,
                     name='ManualTracking'
                 )
         except:
             project = Project(
-                org=Org.objects.get(id=org),
```
nickolashe: no need for an extra db call

nickolashe:
@karbmk any other changes needed here in your opinion?

nickolashe:
@karbmk am i wrong or it looks like you've been merging Tim's changes directly off his branch? if so, lets only be merging develop while working on features in parallel

openCurrents/forms.py
```diff
@@ -232,6 +239,82 @@ class EventRegisterForm(forms.Form):
...
+
+    choices_init = [("select_org","Select organization")]
+    choices = [
+        (org.id, org.name)
+        for org in Org.objects.all().order_by('name')
+    ]
+    choices = choices_init + choices
+    org = forms.ChoiceField(
+        choices=choices,
+        widget=forms.Select(attrs={
```
nickolashe: @dannypernik which css style format should we use for a select box?

dannypernik: @nickolashe the select here has id=user-orgs. I'm going to remove the two-margin-bottom class and add that to the id

dannypernik: @nickolashe hm there's actually no css tied to that id. The css is specified for all select fields, and the width is determined by the columns of a parent div

dannypernik: But I'll go ahead and move the margin to the id in app.css

nickolashe:
@dannypernik i was asking which class are we using for a select box like this? this is where we'd need to specify it

openCurrents/forms.py
nickolashe: i was asking which class are we using for a select box like this? this is where we'd need to specify it

nickolashe: not following the discussion about the id

dannypernik:
https://github.com/opencurrents/opencurrents/blob/master/openCurrents/static/css/app.css#L226 all select boxes are styled the same. In my latest PR I added margin to the id="user-orgs" instead of using the two-margin-bottom class. This way only the id will need to be defined in forms.py.

nickolashe:
why is the id approach preferred over specifying the class?

dannypernik:
Best practice says to keep styles in .css file. I've made a bunch of classes for things like margins (especially margins) so that I don't have to go to the .css file every time. However, I'd much rather go to the css file to make this change than go to forms.py.

nickolashe:
we can implememt our custom select widget using css class and then there is no need to change anything in forms.py

nickolashe:
so which class are we using for selects again?

dannypernik:
None. We are styling select elements directly using the CSS I linked to above. My PR includes an update to the CSS of the ID.

nickolashe:
@dannypernik having to change css of the id does not seem scalable to me. thoughts?

dannypernik:
@nickolashe IDs are used for unique elements. There is no reason to expect that all select boxes should have a bottom margin of 2rem.

nickolashe:
@dannypernik ok, so the bottom line is we now specify styling in app.css and not worry about classes?

nickolashe:
@dannypernik @karbmk somehow we ended up with a conflict in time-tracker. any idea how?

dannypernik:
@nickolashe: @karthik has an idea about the conflict, he'll get back to us on that.
As for styling, if you see a class in an html element, we are using it for styling. Going forward I actually have no problem putting classes into forms.py, I didn't need to complicate things.

karthik:
@dannypernik You keep tagging the wrong person.

nickolashe:
@dannypernik did you get my comment re: making a custom input widget?

nickolashe:
i've created an issue on this topic a while back:
https://github.com/opencurrents/opencurrents/issues/30

openCurrents/forms.py
dannypernik: @nickolashe if there's no class in the html, we don't need a class.

dannypernik:
@nickolashe commented on #30

PR #161, opened 2017-07-12
Event type in model

karbmk:
Event type field added into models


openCurrents/models.py
```diff
@@ -145,6 +145,17 @@ class Event(models.Model):
...
+    MANUAL = 'MN'
+    GROUP = 'GR'
+    OTHER = 'OT'
+    event_type_choices = (
+        (MANUAL, 'ManualTracking'),
+        (GROUP, 'Group'),
+        (OTHER, 'Other'),
+    )
+    event_type = models.CharField(max_length=2,
+                                      choices=event_type_choices,
```
nickolashe: indentation looks off

openCurrents/models.py
```diff
@@ -145,6 +145,17 @@ class Event(models.Model):
     # coordinator contact info
     coordinator_firstname = models.CharField(max_length=128)
     coordinator_email = models.EmailField()
+    MANUAL = 'MN'
+    GROUP = 'GR'
+    OTHER = 'OT'
```
nickolashe: any thoughts what we would use this for?

openCurrents/views.py
```diff
@@ -226,6 +226,7 @@ def track_hours(self, form_data):
         event = Event(
             project=self.project,
             description=form_data['description'],
+            event_type="MN",
```
nickolashe: is this the only place where we are creating events?

openCurrents/models.py
karbmk: Yep this is how its filtered in the templates. They are basically the values of the tag.

openCurrents/models.py
karbmk: Got it. Will make it proper.

openCurrents/views.py
karbmk: The manual events. Yes. If they are being created elsewhere I have made GROUP the default. So it shouldn't be a problem.

openCurrents/models.py
nickolashe: @karbmk i meant the 'OT' type specifically. i think lets leave it out until we have other types to deal with.

openCurrents/views.py
nickolashe: got it, that will work

openCurrents/models.py
karbmk: Oh sorry. I thought all of it. I will do the changes.

openCurrents/models.py
```diff
@@ -154,8 +154,8 @@ class Event(models.Model):
         (OTHER, 'Other'),
     )
     event_type = models.CharField(max_length=2,
-                                      choices=event_type_choices,
-                                      default=GROUP)
+                                  choices=event_type_choices,
```
nickolashe: @karbmk we are using a slightly different practice for breaking lines and indenting, see this for an example:
https://github.com/opencurrents/opencurrents/blob/master/openCurrents/views.py#L312-L316

nickolashe:
@karbmk you've tested this locally, correct?

karbmk:
Yes

nickolashe:
ok, feel free to merge one the indentation is consistent with the rest of the codebase

karbmk:
Ok sure I will do that

PR #162, opened 2017-07-13
Multiple small changes, please see comments

tsiology:
- live-dashboard now prepends new volunteer sign ins to the top of the list.

- profile page will now display welcome popup only on the first visit through use of browser localStorage.

- process_resend now redirects with a status_msg included. status_msg displays after Resend is hit.
      _DONE: status success/fail flag passed to template instead of entire status_msg_

- sendContactEmail added to alert coordinators of new user registrations to their event. Optional contact message will be included in email if provided.
      _DONE: contact email sends with subject line_

- DatetimeEncoder allows for serialization of datetime from sendContactEmail to mandrill template.

### DIFF
--- a/openCurrents/views.py
+++ b/openCurrents/views.py
@@ -55,9 +55,7 @@ def diffInHours(t1, t2):
 class DatetimeEncoder(json.JSONEncoder):
     def default(self, obj):
         if isinstance(obj, datetime):
-            return obj.strftime('%Y-%m-%dT%H:%M:%SZ')
-        elif isinstance(obj, date):
-            return obj.strftime('%Y-%m-%d')
+            return obj.strftime('%m-%d-%Y')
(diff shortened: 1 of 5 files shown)


openCurrents/views.py
```diff
@@ -604,13 +614,108 @@ def event_register(request, pk):
     if form.is_valid():
         user = request.user
         event = Event.objects.get(id=pk)
+        message = form.cleaned_data['contact_message']
+        
+        #check for existing registration
+        event_records = UserEventRegistration.objects.filter(user=user, event=event, is_confirmed=True).exists()
```
nickolashe: @tsiology it is better to use id's for filtering since those are indexed columns

nickolashe:
@tsiology looks good to me, just a small change requested above

PR #163, opened 2017-07-14
Approve hours

karbmk:
Able to get the data from models and post the user input:
1. Admin will be able see the hours recorded manually by volunteers.
2. As of now only the data from last week from now could be seen in the approve-hours page.


karbmk:
Sure. Do let me know if there are any concerns or issues.

openCurrents/templates/approve-hours.html
```diff
@@ -13,89 +13,108 @@
...
-                  <p>Thursday<span>3 hrs</span></p>
-                  <p class="edit-volunteer center">
-                    <input type="radio" id="defer-vol-1" name="vol-status-1" class="vol-radio hidden"/>
-                      <label for="defer-vol-1">
+        <form action ="" method="post">
+          {% csrf_token %}
+          <div class="row left one-margin-top one-margin-bottom">
+            <div class="small-12 medium-8 small-centered columns">
+              <ul id="volunteer-list" class="accordion" data-accordion data-allow-all-closed="true">
+                {% for key, value in users.items %}
```
nickolashe: users.iteritems

openCurrents/views.py
```diff
@@ -112,8 +115,63 @@ def get_context_data(self, **kwargs):
...
+
+    def get_queryset(self):
+        time_log = OrderedDict()
+        keys = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday', 'Total']
+        value = 0
+        items = [(key, value) for key in keys]
+        all_volunteer_data = UserTimeLog.objects.all()
+        #print(all_volunteer_data[0].is_verified)
+        for i in all_volunteer_data:
+            if (str(i.datetime_start).split(" ")[0] == str(i.datetime_end).split(" ")[0]) and i.is_verified != True:
```
nickolashe: @karbmk are you checking for same day here? why not use `datetime` operations instead of operating with strings?

nickolashe:
@karbmk some of the changes here are challenging to follow. can you please include more comments with your code?

openCurrents/views.py
nickolashe: also remember the times are going to be in UTC timezone

openCurrents/views.py
```diff
@@ -112,8 +115,63 @@ def get_context_data(self, **kwargs):
...
+        post_data = self.request.POST['post-data']
+        templist = post_data.split(",")
+        templist[:] = [item for item in templist if item != '']
+        try:
+            for i in templist:
+                user = User.objects.get(username=i)
+                time_log = UserTimeLog.objects.filter(user=user).update(is_verified = True);
+            return redirect('openCurrents:hours-approved')
+        except:
+            return redirect('openCurrents:approve-hours')
```
nickolashe: why are we redirecting to the same page here?

nickolashe:
@karbmk I don't see any hours on the approve-hours page. I checked some users in for this week on live-dashboard page though.

openCurrents/views.py
karbmk: Since the approve of hours did'nt happen. I thought redirecting back. Is there any other page i should redirect to?

openCurrents/views.py
karbmk: I will change that

openCurrents/views.py
```diff
@@ -112,8 +115,63 @@ def get_context_data(self, **kwargs):
...
+class ApproveHoursView(LoginRequiredMixin, SessionContextView, ListView):
     template_name = 'approve-hours.html'
+    context_object_name = 'users'
+
+    def get_queryset(self):
+        time_log = OrderedDict()
+        keys = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday', 'Total']
+        value = 0
+        items = [(key, value) for key in keys]
+        all_volunteer_data = UserTimeLog.objects.all()
```
karbmk: Do you mean the usertimelog objects??

openCurrents/views.py
nickolashe: yes, error 500 page

nickolashe:
@karbmk I am seeing a few issues

1. `datetime_end` is not guaranteed to exist, so this change throws an error every time it is accessed
2. even with `datetime_end`, get_hours, line 174 throws an error: `invalid literal for int() with base 10: '28.283518'`

karbmk:
Hi @nickolashe , if I understand it correctly the filtering of date should be for the week before from today right? Also how about organisations. Is it the organisation of the admin? I din't quite get it. Can you please tell me more on this.

karbmk:
Also, The datetime_end will exist since the volunteer won't be allowed to submit his hours if he doesn't have a end time mentioned.

openCurrents/views.py
```diff
@@ -171,7 +170,7 @@ def post(self, request):
 
     def get_hours(self, time_str):
         h, m, s = time_str.split(':')
-        return int(h) + int(m)/60 + int(s)/3600
+        return float(h) + float(m)/60 + float(s)/3600
```
nickolashe: I think we need to round this figure to nearest 15 min interval

nickolashe: I am currently seeing 2.00785653278 hrs

nickolashe:
@karbmk Yes, we want to filter usertimelogs by:

1. org_id
2. datetime_start

since we only want to display the times for the logged in admin (org) and within a certain timeframe

nickolashe:
@karbmk `datetime_end` does not always exist for event based checkins, but since you are filtering by `is_verified` as well i think we happen to be good here. I'd filter by manual as well just to be sure. Btw, you can chain filters using `.` operator.

nickolashe:
See this for an example of chaining queryset filters:
https://github.com/opencurrents/opencurrents/blob/master/openCurrents/views.py#L224-L228

karbmk:
Yep I tried them chaining and it worked. I will add the manual filter as well. Can you take a look on filters in https://github.com/opencurrents/opencurrents/pull/163/commits/cac368024f0a5eaae134b51fd82749a410202587 and provide feedback

openCurrents/views.py
```diff
@@ -125,13 +125,23 @@ def get_queryset(self):
...
         userid = self.request.user.id
-        user = User.objects.get(id=userid)
-        org = OrgUser.objects.get(user=user).org
-        project = Project.objects.filter(org=org)
-        events = Event.objects.filter(project=project)
+        #user = User.objects.get(id=userid)
+        orgid = OrgUser.objects.get(user__id=userid).org.id
+        project = Project.objects.filter(org__id=orgid)
+        events = Event.objects.filter(
+            project=project
```
nickolashe: `project__id=project.id`

nickolashe: oh, looks like `project` is actually a list? so lets rename to `projects`

karbmk: But It's rendering multiple projects

karbmk: Got it. I will rename it

openCurrents/views.py
```diff
@@ -125,13 +125,23 @@ def get_queryset(self):
...
+        ).filter(
+            event_type='MN'
+        )
         #date_today = datetime.now()
         last_week = datetime.now() - timedelta(days=7)
-        all_volunteer_data = UserTimeLog.objects.filter(is_verified=False).filter(event=events).filter(datetime_start__gte = last_week)#all()
+        all_volunteer_data = UserTimeLog.objects.filter(
```
nickolashe: @karbmk @tsiology thought you'll find this example useful. the order of operations is actually important here if we want this thing to scale up. here is the schema:
`
sqlite> .schema openCurrents_usertimelog
CREATE TABLE "openCurrents_usertimelog" ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "datetime_start" datetime NOT NULL, "is_verified" bool NOT NULL, "date_created" datetime NOT NULL, "date_updated" datetime NOT NULL, "user_id" integer NOT NULL REFERENCES "auth_user" ("id"), "event_id" integer NOT NULL REFERENCES "openCurrents_event" ("id"), "datetime_end" datetime NULL);
CREATE INDEX "openCurrents_usertimelog_e8701ad4" ON "openCurrents_usertimelog" ("user_id");
CREATE INDEX "openCurrents_usertimelog_4437cfac" ON "openCurrents_usertimelog" ("event_id");
`

notice there is an index on the event id, which means querying by events is fast. so i would filter in the following order:
1. event
2. datetime
3. is_verified

openCurrents/views.py
```diff
@@ -126,7 +126,9 @@ def get_queryset(self):
         items = [(key, value) for key in keys]
         userid = self.request.user.id
         #user = User.objects.get(id=userid)
-        orgid = OrgUser.objects.get(user__id=userid).org.id
+        org = OrgUser.objects.filter(user__id=userid)
```
nickolashe: we should be able to get `orgid` from the context (retrieved by get_context_dat as on line 108)

openCurrents/views.py
karbmk: Oh that's an interesting detail. Thanks for letting me know. I will make the necessary changes.

openCurrents/views.py
nickolashe: @karbmk have you checked that this actually works as intended with a list a query param?

nickolashe:
@karbmk I looked at this today and made some changes. I think more work is need on this pr.

openCurrents/views.py
```diff
@@ -155,52 +151,69 @@ def get_queryset(self):
...
-                del time_log[i]
+
+        time_log = OrderedDict([
+            (k, time_log[k])
+            for k in time_log
+            if time_log[k]['Total'] > 0
+        ])
         return time_log
 
     def post(self, request):
```
nickolashe: @karbmk from what i understand, this will verify all of the existing hours for the volunteer?

nickolashe:
@karbmk most importantly, i don't think anything happens if the hours are declined. i think we need to go ahead and delete those records in usertimelog table in that case.

nickolashe:
@dannypernik @karbmk also, do you think we should support pagination by week? with the current implementation, if hours are not approved within a week they will never be presented for approval again.

nickolashe:
also, although never presented, they will be nonetheless approved once any subsequent hours are approved (see https://github.com/opencurrents/opencurrents/pull/163#discussion_r129711317)

dannypernik:
@nickolashe @karbmk agreed we need weekly pagination. I don't think the template needs to change, just the dates on the page

nickolashe:
@dannypernik what about past / forward scroll buttons?

openCurrents/views.py
```diff
@@ -155,52 +151,69 @@ def get_queryset(self):
...
-                    time_log[str(i.user)]['Total'] += rounded_time_mdv2
+        )
+
+        time_log = OrderedDict()
+        items = {'Total': 0}
+
+        for timelog in eventtimelogs:
+            user_email = timelog.user.email
+
+            # check if same day and duration longer than 15 min
```
nickolashe: @karbmk lets try to have a brief comment preceding each code block explaining what it does?

dannypernik:
@nickolashe thinking that you proceed to next week after clicking "approve"

dannypernik:
Will need to add a status message to page to make it clear that hours were approved

openCurrents/views.py
karbmk: Yes

openCurrents/views.py
karbmk: Yep. Point noted

openCurrents/views.py
```diff
@@ -132,9 +137,212 @@ def get_context_data(self, **kwargs):
...
+        week_startdate = oldest_timelog.datetime_start
+        week_startdate_monday = week_startdate - timedelta(days=week_startdate.weekday())
+        today = timezone.now()
+        #print(week_startdate_monday)
+
+        # loop with 7 day increments from monday before oldest timelog until today
+        # building the list weeks with all timelogs per week
+        #while( week_startdate_monday <= today ):
+        time_log_week = OrderedDict()
+        for x in range(0,1):
```
nickolashe: @karbmk what's the reason behind this?

openCurrents/views.py
```diff
@@ -132,9 +137,212 @@ def get_context_data(self, **kwargs):
...
+                  
+                    #build manual tracking filter, currently only accessible by OrgUser...  
+                    userid = user.id
+                    org = OrgUser.objects.filter(user__id=userid)
+                    orgid = org[0].org.id
+                    projects = Project.objects.filter(org__id=orgid)
+                    events = Event.objects.filter(project__in=projects).filter(event_type='MN')
+
+                    #check if the volunteer is declined and delete the same
+                    if i.split(':')[1] == '0' and i != '':
```
nickolashe: @karbmk i think if the first check succeeds, it warrants the second one already

openCurrents/views.py
```diff
@@ -132,9 +137,212 @@ def get_context_data(self, **kwargs):
...
+                #split the data for user, flag, and date info
+                try:
+                    user = User.objects.get(username=i.split(':')[0])
+                    week_date = datetime.strptime( i.split(':')[2], '%m-%d-%Y')
+                  
+                    #build manual tracking filter, currently only accessible by OrgUser...  
+                    userid = user.id
+                    org = OrgUser.objects.filter(user__id=userid)
+                    orgid = org[0].org.id
+                    projects = Project.objects.filter(org__id=orgid)
```
nickolashe: @karbmk main reason why this PR is not working for me is the following:
1. when a volunteer submits manual hours, they are free to pick any org
2. we then create a project / event under that org, and a usertimelog linked to that project / event
3. these project / events would then be missed by this lookup here, because you are always using a fixed orgid (that of the user)

openCurrents/views.py
karbmk: @tsiology was the one who coded that. He might be the better person to answer. But I remember saying that it was for the pagination if i am not wrong.

openCurrents/views.py
karbmk: Got it. I will check for empty and inside that check for '0'

openCurrents/views.py
karbmk: I think i got it. I am working on it.

karbmk: While creating the tracking hours i have already made sure that the steps you told are implemented. I will now implement the changes corresponding to it in approve hours.

openCurrents/views.py
tsiology: @nickolashe @karbmk The implementation underwent some changes that have not been cleaned up. Originally, a list was used to send all UserTimeLogs in 7 day sets. This was changed to send only the oldest weeks worth, one week at a time. That's why the while loop was replaced with a for(0,1), I need to clean it up since no longer needed

nickolashe:
@karbmk @tsiology I went ahead and merge this as is, but lets find some time to discuss using `JSON` when posting data for views

PR #174, opened 2017-07-19
Invite volunteers latest

karbmk:
Invite volunteers module added


karbmk:
@nickolashe check this branch for invite volunteers

openCurrents/views.py
```diff
@@ -487,9 +488,55 @@ class ProjectDetailsView(TemplateView):
...
-class InviteVolunteersView(TemplateView):
+class InviteVolunteersView(LoginRequiredMixin, SessionContextView, TemplateView):
     template_name = 'invite-volunteers.html'
 
+    def post(self, request, *args, **kwargs):
+        userid = self.request.user.id
+        user = User.objects.get(id=userid)
+        post_data = self.request.POST
+        k = []
+        Organisation = OrgUser.objects.get(user=user).org.name
```
nickolashe: @karbmk lets do all lookups by id, so `user__id=user.id` in this case

karbmk: Got it. I will do that. The form email does get sent. But i will take a look at it once again.

karbmk:
Hi @nickolashe, I have changed the filtering of Org name and also the issue of form email being sent. the issue with sending email was the patch I took from the old branch had an input tag different which made it not work. I have it rectified. Please take a look at it and let me know if there are any other issues.

openCurrents/views.py
```diff
@@ -496,7 +496,7 @@ def post(self, request, *args, **kwargs):
         user = User.objects.get(id=userid)
         post_data = self.request.POST
         k = []
-        Organisation = OrgUser.objects.get(user=user).org.name
+        Organisation = OrgUser.objects.get(user__id=userid).org.name
```
nickolashe: @karbmk FYI, this will raise a `MultipleObjectsReturned` for any user that is affiliated to multiple orgs.

openCurrents/views.py
```diff
@@ -496,7 +496,9 @@ def post(self, request, *args, **kwargs):
         user = User.objects.get(id=userid)
         post_data = self.request.POST
         k = []
-        Organisation = OrgUser.objects.get(user__id=userid).org.name
+        OrgUsers = OrgUser.objects.filter(user__id=userid)
```
nickolashe: same comment as https://github.com/opencurrents/opencurrents/pull/163/commits/e500cdef784565082f346f32398cfbe8d2b49b1a#r128626383

nickolashe:
@karbmk FYI, `volunteers-invited` still says "X volunteers invited". can we pass in the actual number to the template? or change the message otherwise.

karbmk:
I will take a look at it. not a problem

nickolashe:
CreateEventView needs to redirect to invite-volunteers

openCurrents/views.py
```diff
@@ -935,17 +955,53 @@ def process_signup(request, referrer=None, endpoint=False, verify_email=True):
...
+                    return HttpResponse(user.id, status=200)
+
+                if user.has_usable_password():
+                    logger.info('user %s already verified', user_email)
+                    return redirect(
+                        'openCurrents:signup',
+                        status_msg='User with this email already exists'
+                    )
+                else:
+                    logger.info('user %s has not been verified', user_email)
```
nickolashe: it does not look like we are returning any response at all in this case

nickolashe:
@karbmk I am merging this due to time constraints but lets find some time to discuss in more detail

PR #177, opened 2017-07-20
Password reset

tsiology:
New password reset functionality can be requested from login page. If user already has a valid password, they will be emailed a link to the reset-password template.


nickolashe:
Can we fix the links and add the token to the url pls

tsiology:
@nickolashe Fixed redirections and added token verification in process_reset_password.

nickolashe:
@tsiology thanks, will give this a try and let you know

nickolashe:
@tsiology looks good

PR #178, opened 2017-07-24
Export data

karbmk:
Export data modle


nickolashe:
@karbmk can you please use more comments as well as provide a short summary with each pr?

openCurrents/views.py
```diff
@@ -131,9 +132,74 @@ class CausesView(TemplateView):
...
+        writer.writerow(k_dict.keys())
+        for i in vol_personal_info:
+            usertimelog_info = UserTimeLog.objects.filter(user=i)
+            for j in usertimelog_info:
+                try:
+                    datetime_duration = j.datetime_end-j.datetime_start
+                except:
+                    datetime_duration = '00:00:00'
+                if post_data['start-date'] != u'':
+                    if (str(j.event.datetime_start)> str(post_data['start-date']) or str(j.event.datetime_start)==str(post_data['start-date'])) \
```
nickolashe: @karbmk why are we using string comparison on datetimes?

karbmk: request data was coming as unicode and the datetime both of them had different offsets. So thought of comparing strings. I will change it to datetime.

nickolashe:
lets modify the logic of calculating durations:
1. end time may be absent
2. start / end time may be wrong (we go by event start / end times)
3. enforce constraint of 1 row per user per event
4 (optional). extract the existing logic from profile view that calculates user balance in currents based of usertimelog

openCurrents/views.py
```diff
@@ -208,17 +208,17 @@ def post(self, request):
                     s_dt_db.replace(tzinfo=utc)==datetime.strptime(s_dt_ui, '%Y-%m-%d').replace(tzinfo=utc) ) \
                     and (e_dt_db.replace(tzinfo=utc) <datetime.strptime(e_dt_ui, '%Y-%m-%d').replace(tzinfo=utc)  or\
                     e_dt_db.replace(tzinfo=utc) ==datetime.strptime(e_dt_ui, '%Y-%m-%d').replace(tzinfo=utc) ):
-                        cleaned_list = [str(i.first_name), str(i.last_name), str(i.email), str(datetime_duration), str(j.datetime_start),\
-                            str(j.datetime_end), str(j.event.datetime_start), str(j.event.location), str(j.event.project.name)]
+                        cleaned_list = [str(i.first_name), str(i.last_name), str(i.email), str(datetime_duration), str(j.event.datetime_start),\
```
nickolashe: @karbmk you can python `map` function to apply `str()` conversion to the entire list

nickolashe:
@karbmk what about the uniqueness constraint (point 3 above)?

nickolashe:
bump

nickolashe:
@karbmk there are some conflicts here as well

openCurrents/views.py
```diff
@@ -384,7 +384,7 @@ def post(self, request):
         writer = csv.writer(response)
         writer.writerow(k_dict.keys())
         for i in vol_personal_info:
-            usertimelog_info = UserTimeLog.objects.filter(user=i).filter(is_verified=True)
+            usertimelog_info = UserTimeLog.objects.filter(user=i,event__in=event_info).filter(is_verified=True)
```
nickolashe: @karbmk i am not sure i am seeing how this addresses the unique user per event constraint. just to be clear, by design a given user can have many `UserTimeLog` records corresponding to the same event, but we want only one entry in the exported data.

karbmk: What if the event is ManualTracking(A user can have multiple ManualTracking events for different days but user and event will be the same). Also if it's even other event should we display the first occurrence of the user and event or how should it display?

nickolashe: i thought we are creating a separate event for each MT?

nickolashe: for group events we should just display event start and end times since this is what we use to assign currents. @dannypernik do you agree?

karbmk: Yeah sorry I confused project with event. Event is created for every manually tracked hours. I think that the uniqueness of user to event is maintained. I might be missing something here. I am not able to follow the **unique user per event constraint**. Can you provide me an example fo it please?

karbmk: > for group events we should just display event start and end times since this is what we use to assign currents. @dannypernik do you agree?
Does this mean that we are not going to display other fields for those records??

nickolashe: sorry for the confusion, no that does not mean we are not displaying other records, we are going to display other records as well

nickolashe: @karbmk i meant for group events a volunteer can be checked in and out many times, right?

karbmk: Ah!! got it. Basically all the manual tracking records and group event trackings record should be only one record with start and end time same as that of event right?

dannypernik: @karbmk @nickolashe correct, we assume that a volunteer checked in at an event is present for the full duration. Time of check-in during event is irrelevant.

nickolashe: @karbmk yes, that's correct

karbmk: Got it. I will work towards that.

openCurrents/views.py
```diff
@@ -184,8 +186,14 @@ def post(self, request):
                 except:
                     datetime_duration = '00:00:00'
                 if post_data['start-date'] != u'':
-                    if (str(j.event.datetime_start)> str(post_data['start-date']) or str(j.event.datetime_start)==str(post_data['start-date'])) \
-                    and (str(j.event.datetime_end)<str(post_data['end-date']) or str(j.event.datetime_end)==str(post_data['end-date'])):
+                    s_dt_db = j.event.datetime_start
+                    s_dt_ui = post_data['start-date']
+                    e_dt_db = j.event.datetime_end
+                    e_dt_ui = post_data['end-date']
+                    if (s_dt_db.replace(tzinfo=utc) > datetime.strptime(s_dt_ui, '%Y-%m-%d').replace(tzinfo=utc)  or\
```
nickolashe: @karbmk why are we replacing timezones here? i think the one from the DB is going to be in utc, while the one from the UI is local.

nickolashe: also, is not the above equivalent to using `>=` comparison?

openCurrents/views.py
```diff
@@ -184,8 +186,14 @@ def post(self, request):
...
+                    s_dt_db.replace(tzinfo=utc)==datetime.strptime(s_dt_ui, '%Y-%m-%d').replace(tzinfo=utc) ) \
+                    and (e_dt_db.replace(tzinfo=utc) <datetime.strptime(e_dt_ui, '%Y-%m-%d').replace(tzinfo=utc)  or\
```
nickolashe: similarly, is not this equivalent to `<=` ?

nickolashe: finally, why do we need to do this comparison at all?

openCurrents/views.py
karbmk: I think i need to remove the replace and think of different comparision!

openCurrents/views.py
karbmk: The comparison is to make sure that the event start and end time matches the user defined time in the UI.

openCurrents/views.py
```diff
@@ -431,7 +430,7 @@ def post(self, request):
                             writer.writerow(cleaned_list)#write to the CSV file
                 else:
                     #if the user input in start-time is empty
-                    if (str(j.event.datetime_end)<str(post_data['end-date']) or str(j.event.datetime_end)==str(post_data['end-date'])):
+                    if str(j.event.datetime_end) <= str(post_data['end-date']):
```
nickolashe: lets always use datetime comparison when comparing dates

openCurrents/views.py
```diff
@@ -374,6 +374,7 @@ def post(self, request):
             ('event-name',8)
         ])
         utc=pytz.UTC
+        tz = 'America/Chicago'
```
nickolashe: what's the reason behind hardcoding this? can we just set to the timezone of the org, where the org is the first one in the list of userorg affiliations?

openCurrents/views.py
```diff
@@ -416,10 +417,8 @@ def post(self, request):
                     s_dt_ui = post_data['start-date']
                     e_dt_db = j.event.datetime_end
                     e_dt_ui = post_data['end-date']
-                    if (s_dt_db.replace(tzinfo=utc) > datetime.strptime(s_dt_ui, '%Y-%m-%d').replace(tzinfo=utc)  or\
-                    s_dt_db.replace(tzinfo=utc)==datetime.strptime(s_dt_ui, '%Y-%m-%d').replace(tzinfo=utc) ) \
-                    and (e_dt_db.replace(tzinfo=utc) <datetime.strptime(e_dt_ui, '%Y-%m-%d').replace(tzinfo=utc)  or\
-                    e_dt_db.replace(tzinfo=utc) ==datetime.strptime(e_dt_ui, '%Y-%m-%d').replace(tzinfo=utc) ):
+                    if ( s_dt_db.astimezone(pytz.timezone(tz)) >= pytz.timezone(tz).localize(datetime.strptime(s_dt_ui, '%Y-%m-%d')) ) \
```
nickolashe: the datetimes coming out of the db are already tz-aware, so no reason to convert those to any other timezone for comparison purposes

nickolashe: for the end date, should not we include all of the events from the day specified by the user in the ui?

nickolashe: datetime.strptime(e_dt_ui, '%Y-%m-%d') will evaluate to the beginning of that day from what i understand

openCurrents/views.py
karbmk: Since we are just comparing and not saving any of the data. I needed a timezone so that they could be compared.

openCurrents/views.py
karbmk: But without the timezone conversion it gives ```cannot compare naive tz with "some other" tz``` error

nickolashe: can you include a snapshot? i mean we still need to convert the `_ui` timezones to tz-aware, just no conversion at all for `_db` timezones needed i think

openCurrents/views.py
nickolashe: @karbmk tz-aware do not need to be in the same timezone for comparing. `_db` times are in UTC, and `_ui` should be localized to the org's timezone. they can then be compared directly.

openCurrents/views.py
karbmk: Ok i will upload a snapshot.

karbmk: screen shot 2017-08-16 at 3 03 44 pm

nickolashe: ok, thanks. can you push the change that produced this snapshot so i can test please?

karbmk: Ok sure

karbmk: I have committed the code with error

openCurrents/views.py
```diff
@@ -525,11 +525,12 @@ def post(self, request):
                 if post_data['start-date'] != u'':
                     #if the user fill the start-time this condition is executed
                     s_dt_db = j.event.datetime_start
+                    logger.info(s_dt_db.tzinfo)
```
nickolashe: this shows that db times are tz-aware

openCurrents/views.py
```diff
@@ -525,11 +525,12 @@ def post(self, request):
...
                     s_dt_ui = post_data['start-date']
                     e_dt_db = j.event.datetime_end
                     e_dt_ui = post_data['end-date']
-                    if ( s_dt_db >= datetime.strptime(s_dt_ui, '%Y-%m-%d') ) \
-                    and ( e_dt_db <= datetime.strptime(e_dt_ui, '%Y-%m-%d') ):
+                    if ( s_dt_db >= pytz.timezone(tz).localize(datetime.strptime(s_dt_ui, '%Y-%m-%d')) ) \
```
nickolashe: times from the form need to be localized, however

nickolashe: @karbmk sorry if this was confusing, can you let me know if this makes sense now (check that these changes are working)

nickolashe: once you are happy with them, we can remove the logging statement above

nickolashe: we still need to fetch the true timezone of the org instead of a hardcoded one on line 488

karbmk: Ahh got it!!
Regarding the line 488 how could we get the org timezone? Do we do that for every user inside the loop??

nickolashe: no, just the admin's org. this feature is intended for org admins, correct? @dannypernik

dannypernik:
@karbmk what's the status here?

dannypernik:
@nickolashe I've just resolved conflicts in this PR. For all conflicts except one, I chose the version in develop. The one exception was ExportHoursView, where develop branch didn't have anything beyond declaring the template.

Also note that I've changed a couple of the names of the checkbox inputs in export-hours. Not sure if that has an impact on implementation.

nickolashe:
@dannypernik sounds good, thanks for taking the lead. how are the use cases coming along?

dannypernik:
@nickolashe use cases here: https://docs.google.com/spreadsheets/d/15mnrkl4Rc5dsFX8VAGOeuPPfCQTPHyOWV1SVwkXIqNQ/edit#gid=1212779018

nickolashe:
i think this looks good, thanks for completing those @dannypernik

nickolashe:
@dannypernik can we evaluate the current state of this PR, i.e. is the data exported at all, if so which fields are being exported?

dannypernik:
@nickolashe the "export" button on /export-data doesn't do anything. I'll stop making ignorant comments without testing.

nickolashe:
ok, i will take a look. any ideas why the button is not doing anything yet?

dannypernik:
Not for the first time, I used `<a type="submit">` which is not valid. I looked at that specifically yesterday and still got it mixed up. type=submit only works with `<button>` or `<input>` elements. I've pushed the fix. However, immediately got this error when submitting. Same error with or without start date (both are happy paths):

nickolashe:
@dannypernik that looks like a bug in the code. i'll take a look.

PR #179, opened 2017-07-24
Admin currents after volunteer checkin

tsiology:
Event admin/coordinator will be awarded currents equal to the hours of the event when the first volunteer is checked in for that event.


openCurrents/views.py
```diff
@@ -611,6 +612,19 @@ def event_checkin(request, pk):
         )
 
         if checkin:
+            # award admin/coordinator currents on checkin only if not already awarded
+            if not event.admin_awarded:
+                admin = User.objects.get(email=event.coordinator_email)
```
nickolashe: i think we want to award the user using the live-dashboard in the current session rather than coordinator_email, which may be absent

openCurrents/views.py
```diff
@@ -611,6 +612,19 @@ def event_checkin(request, pk):
...
+                account = Account.objects.get(user=admin)
```
nickolashe: user__id

openCurrents/views.py
```diff
@@ -611,6 +612,19 @@ def event_checkin(request, pk):
...
+                event.admin_awarded = True
+                event.save()
+
+                hours = decimal.Decimal(diffInHours(event.datetime_start,event.datetime_end))
+                account.amount += hours
+                account.save()
+
+                logger.warning('Awarded admin %s hours: %s currents for event id: %s', admin, hours, pk)
```
nickolashe: I think this would be a debug or info level log, not a warning

openCurrents/views.py
nickolashe: or incorrect

tsiology: @nickolashe please see updates using request.user.id passed from the live-dashboard template and user__id to get Account object. Thanks!

nickolashe:
https://github.com/opencurrents/opencurrents/pull/179

tsiology:
The latest commit simplifies the logic for verifying the first check in. On check in, if no UserTimeLog for the coordinator for this event exists, create one. All other checkins will see this is already created. Verified only one admin UserTimeLog is being created after multiple checkins.

Still need to verify currents showing for the admin on their profile page.

tsiology:
@nickolashe this PR now adds two migration files, one to add admin_awarded field to Events, and one to remove it. Is it safe/practical to just remove both of these files since models.py is unchanged?

nickolashe:
@tsiology yes, i think so

openCurrents/views.py
```diff
@@ -620,6 +621,16 @@ def event_checkin(request, pk):
         )
 
         if checkin:
+            # create admin/coordinator UserTimeLog on checkin only if not already done
+            if not UserTimeLog.objects.filter(event__id=event.id, user__id=request.user.id):
```
nickolashe: @tsiology this looks good, however one question. do you see any problem here if this code were to run concurrently (i.e. we had several threads / nodes processing check in requests)?

openCurrents/migrations/0028_event_admin_awarded.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
nickolashe: i think it is safe to delete both of these migrations as we discussed

openCurrents/views.py
tsiology: @nickolashe  I'm not sure. Would this logic only be a problem if the same admin was logged in on multiple systems? Will the UserTimeLog event/user uniqueness update prevent any problems?

nickolashe: I think the admin would not necessarily need to be logged in on multiple systems, even from a single session checkin requests would be queued to several nodes, then the if clause can simultaneously evaluate to false if they arrive at the same time and the admin will end up getting the credit multiple times

nickolashe: The uniqueness constraint would prevent this, yes

nickolashe: @tsiology for now i think it is totally fine to merge this as is and create an issue to track this going forward

openCurrents/views.py
```diff
@@ -620,6 +621,16 @@ def event_checkin(request, pk):
...
+                usertimelog = UserTimeLog(
+                    user=User.objects.get(id=request.user.id),
+                    event=event,
+                    datetime_start=datetime.now(tz=pytz.UTC)
+                )
+                usertimelog.save()
+    
+
             usertimelog = UserTimeLog(
```
nickolashe: @tsiology I just realized that we need to move the admin checkin after this block of code to avoid double credit in case they decide to check in themselves using live-dashboard

tsiology: That's true. Nice catch.

PR #182, opened 2017-07-25
Create events; Correct current totals

nickolashe:


nickolashe:
@dannypernik Can you review the create events flow?

PR #186, opened 2017-07-25
Migrations

karbmk:
Migration for Event private and public. UserTimeLog duration field added


openCurrents/migrations/0030_auto_20170725_1500.py
```diff
@@ -0,0 +1,25 @@
...
+
+    dependencies = [
+        ('openCurrents', '0029_auto_20170724_1517'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='event',
+            name='is_public',
+            field=models.BooleanField(default=True),
```
nickolashe: @karbmk actually, lets use the default value of `False` as in your initial change. i don't think we want to create any currents without explicit approval.

openCurrents/models.py
```diff
@@ -157,6 +157,8 @@ class Event(models.Model):
         default=GROUP
     )
 
+    is_public = models.BooleanField(default=True)
```
nickolashe: @karbmk is this field used anywhere yet?

openCurrents/models.py
```diff
@@ -218,6 +220,8 @@ class UserTimeLog(models.Model):
     datetime_start = models.DateTimeField('start time')
     datetime_end = models.DateTimeField('end time', null=True, blank=True)
 
+    datetime_duration = models.CharField(max_length=20, default='00:00:00')
```
nickolashe: @karbmk lets make this a decimal field (measured in hours)

karbmk: Ok got it

openCurrents/models.py
karbmk: Not yet

openCurrents/migrations/0029_auto_20170724_1517.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
nickolashe: I think we can delete this migration

PR #227, opened 2017-07-28
Create event start end time status msg

tsiology:
Updated create-event template to grab form ValidationError for Start time after End time.


tsiology:
@nickolashe It looks like this was closed without being merged to develop. Do you know why? I was hoping to follow a similar implementation to display error messages for improper start/end time data in create-event template as was done here.

Issue #228, opened 2017-07-28
live-dashboard: Same user can be checked in to the same event multiple times.


karbmk:
![screen shot 2017-07-28 at 4 12 00 pm](https://user-images.githubusercontent.com/8406712/28736666-8bf7f5e2-73b0-11e7-95cd-bd177e9244cd.png)

karbmk:
Resolved in PR #229

Issue #112, opened 2017-06-15
Opt-in box remains clicked (opted-in) after opting someone in in the live dash board


MaxTattenbach:

nickolashe:
@MaxTattenbach @dannypernik I don't understand what the issue is here

dannypernik:
@nickolashe the thinking here is that once one person opts in, the checkbox is automatically set for future volunteers. @MaxTattenbach 's thinking is that each volunteer should explicitly opt in, otherwise we could have some accidental opt-ins. I tend to agree.

nickolashe:
can you give a specific example perhaps

dannypernik:
Admin adds a new volunteer who wants an invite to openCurrents, they check the "Invite to openCurrents" checkbox. After clicking "Add volunteer", the checkbox remains checked. If the admin doesn't notice this, the next volunteer they add may be accidentally invited to openCurrents.

dannypernik:
<img width="316" alt="screen shot 2017-07-27 at 1 23 37 pm" src="https://user-images.githubusercontent.com/3347095/28706026-ffa6f0de-7337-11e7-9232-feddb7f1d029.png">

@MaxTattenbach @nickolashe what do you think of this as an alternative UI? Not 100% self explanatory but this would mean the admin doesnt have to click the checkbox every time

nickolashe:
not sure i prefer this to the current one

nickolashe:
is the checkbox being reset now each time?

nickolashe:
i just checked that the checkbox is still persistent

PR #229, opened 2017-07-28
Live dashboard checkin

karbmk:
Automatically checks in the volunteer if he/she is checked when the event is live.
Files changed:
1. live-dashboard.html
2. views.py


karbmk:
#228

nickolashe:
lets resolve the conflict pls

nickolashe:
@karbmk @tsiology @dannypernik I am seeing volunteers added to the dashboard checked in on the backend, but the checkbox stays gray. Can you confirm?

nickolashe:
bump

karbmk:
Hi @nickolashe , I verified this. It shows blue tick only when its happening now. Else it will just be added to the usereventregistration.

nickolashe:
@karbmk are you saying that it only checks in the volunteer when the event is happening? or only colors the button so? i think i saw that it checked in a volunteer on my request when the event was not live. can we double check?

karbmk:
Sure

karbmk:
Hi @nickolashe , I checked it. What it basically is doing is add the user in usereventregistration whether it is a live event or an event in the future. But, Will check-in(add in usertimelog) only if it's a live event. If it's an event in the future then it won't create a usertimelog for that user.

nickolashe:
ok, thanks - i'll check again

nickolashe:
@karbmk got it, i think this is correct.

nickolashe:
i think what i meant originally is volunteers can still be checked in by pressing the checkbox for events far in future. but i think this should be a separate issue.

nickolashe:
@karbmk as the last step, do you think you can include changes addressing https://github.com/opencurrents/opencurrents/issues/112 in this PR as well?

karbmk:
Hi Nicko, Sure. If I understand it properly, the opt-in should always be done explicitly. the live-dashboard auto-checkin should be removed. Is that correct?

nickolashe:
@karbmk What do you mean by auto-checkin?

dannypernik:
screen shot 2017-08-09 at 5 05 44 am

@karbmk for clarity, the checkbox mentioned in #112 is the "Invite volunteer to openCurrents" checkbox shown above. Basically after the admin clicks "Add volunteer", the checkbox should reset. Currently, it stays checked.

karbmk:
Oh got it now. I will take a look.

openCurrents/views.py
```diff
@@ -1100,15 +1100,28 @@ def event_register_live(request, eventid):
...
-        user=user,
-        event=event,
-        is_confirmed=True
-    )
-    user_event_registration.save()
-    logger.info('User %s registered for event %s', user.username, event.id)
-
-    return HttpResponse({'userid': userid, 'eventid': eventid}, status=201)
+    user_events = UserEventRegistration.objects.values('user__id','event__id').filter(user__id = userid)
+    user_event_ids = [d for d in user_events if int(userid) == d['user__id'] and int(eventid) == d['event__id']]
```
nickolashe: @karbmk why do we need to build this list when we could add another filter the line above to based on matching eventid?

openCurrents/views.py
```diff
@@ -1100,15 +1100,28 @@ def event_register_live(request, eventid):
...
+        )
+        user_event_registration.save()
+        logger.info('User %s registered for event %s', user.username, event.id)
+    else:
+        logger.info('User %s already registered for event %s', user.username, event.id)
+        return HttpResponse(status=400)
+    tz = event.project.org.timezone
+    event_ds = event.datetime_start.astimezone(pytz.timezone(tz)).time()
+    event_de = event.datetime_end.astimezone(pytz.timezone(tz)).time()
+    d_now = datetime.utcnow().replace(tzinfo=pytz.utc).astimezone(pytz.timezone(tz))
```
nickolashe: @karbmk this does not have to be converted in org's timezone. tz-aware datetimes can be compared directly even if they are in different timezones.

nickolashe:
@karbmk looks good, just a couple of changes based on the comments above

openCurrents/views.py
karbmk: Oh I did'nt know that. Thanks i will change that.

openCurrents/views.py
karbmk: I just need the numbers inside the list but filter is building a list of dictionaries with event__id as key for each element which is sort of not what i wanted.

nickolashe: @karbmk I am not seeing where you are using those numbers though. I am seeing all you are checking for is whether the list is emtpy or not.

karbmk: Hmm, You're right. I thought it was being passed in the URL or something. changing it.

nickolashe:
@karbmk looks good

PR #235, opened 2017-08-02
Confirm account and reset password validation

karbmk:
files modified:
1. forms.py


openCurrents/forms.py
```diff
@@ -105,6 +106,11 @@ def clean(self):
         user_password_confirm = cleaned_data.get('user_password_confirm')
         if user_password and user_password_confirm and user_password != user_password_confirm:
             raise ValidationError(_('Passwords don\'t match'))
+        if re.match(r'((?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?=.*[@#$%]).{8,})', user_password):
```
nickolashe: @karbmk I am pretty sure there is a package that will do this check for us. Can you check please?

nickolashe: bump

karbmk: Ok I will

karbmk:
Hi @nickolashe, these are password validators available in django:
**UserAttributeSimilarityValidator**, which checks the similarity between the password and a set of attributes of the user.
**MinimumLengthValidator**, which simply checks whether the password meets a minimum length. This validator is configured with a custom option: it now requires the minimum length to be nine characters, instead of the default eight.
**CommonPasswordValidator**, which checks whether the password occurs in a list of common passwords. By default, it compares to an included list of 1000 common passwords.
**NumericPasswordValidator**, which checks whether the password isn’t entirely numeric.
I doesn't have 2-3 validation that we need. For this to work we need to define config in settings and write our own validator functions for customisation. So, I was thinking that regex would be a better option. what do you think?

openCurrents/forms.py
```diff
@@ -120,6 +126,11 @@ def clean(self):
         new_password_confirm = cleaned_data.get('new_password_confirm')
         if new_password and new_password_confirm and new_password != new_password_confirm:
             raise ValidationError(_('Passwords don\'t match'))
+        if re.match(r'((?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?=.*[@#$%]).{8,})', user_password):
```
nickolashe: @karbmk FYI, the regex is missing `!`. lets also allow for other special characters

nickolashe:
@karbmk i think the advantage of using django's validators is it would provide a custom error message for each validator. it looks like writing our own validators is really simple:
https://docs.djangoproject.com/en/1.11/topics/auth/passwords/#writing-your-own-validator

nickolashe:
@karbmk this will work for now I think

nickolashe:
can you please double check that the form error messages are shown properly? I am seeing stuff like:
> __all__: Passwords don't match
> user_password_confirm: Ensure this value has at least 8 characters (it has 2).

karbmk:
You mean 2 messages at the same time??

nickolashe:
no, just one at a time. but i think we need to make the messages more user-friendly, like removing the field names

karbmk:
Ok got it.

Issue #213, opened 2017-07-27
live-dashboard: Are all events for all orgs accessible to any admin?


dannypernik:
On my local server, I'm able to change the event id in /live-dashboard/#/ and access any event through a single admin.

dannypernik:
@nickolashe I'm thinking an admin should only be able to access their own org's events, including through URL manipulation.

nickolashe:
agreed

nickolashe:
@tsiology I think we should check whether event org matches that of admin explicitly when serving any pages related to the event

tsiology:
I believe PR #244 now has a working fix so only admins of the org can access live-dashboard of events

Issue #155, opened 2017-07-11
org-signup: Volunteer nominated Keep Austin Beautiful and was designated an admin


dannypernik:
I can only assume this is how tara.rains7@gmail was added as an OrgUser to KAB. Raises the question - if someone names himself a leader or employee of an org, should he or she automatically be assigned admin status?

nickolashe:
Yes, I am aware this is not working yet. This is within the authorization / permissioning scope which I have not had time to get to yet.

nickolashe:
I think they should not be assigned the status right away, only upon approval.

dannypernik:
@nickolashe agreed

nickolashe:
@dannypernik @MaxTattenbach so are we ok with manual approval of admins by us for now?

dannypernik:
Yes

dannypernik:
PR #244

PR #244, opened 2017-08-04
Admin user permissions

tsiology:
Commenting out automatic OrgUser object creation from organization signup methods. OrgUser creation is under review while determining how to create to org admins.


nickolashe:
@tsiology Is this ready for review / testing?

tsiology:
@nickolashe yes, review requested. Simply commenting out OrgUser creations for now

nickolashe:
@tsiology I think I'd like to handle this differently. Lets still record the orguser association, but lets also add an approved field to that table. Anywhere we are checking for an orguser association, we'd also make that conditional on this new field.

nickolashe:
This will allow us to manually review these orguser association requests.

nickolashe:
Note that orguser also an affiliation field, so we should also be checking whether that's equal to `admin` whenever deciding if the user has admin permissions.

tsiology:
@nickolashe I've added back the OrgUser creations but with a new field approved defaulted to False included in the table. I'll need to review all instances of where we filter OrgUser objects and all the check for approved=True. For affiliation=admin, are we ready to start filtering based off that as well?

nickolashe:
@tsiology yep, i think we are

nickolashe:
@tsiology let's also start using django's permissions for the purpose of granting access. please see my last commit.

nickolashe:
as well as this section from the docs:
https://docs.djangoproject.com/en/1.11/topics/auth/default/#permissions-and-authorization

nickolashe:
i.e., once we approve an admin user (manually, for now), they will be granted the `admin` permission

dannypernik:
@nickolashe so an OrgUser will not have an orgid until manually approved? We have {% if orgid %} logic in several places so want to verify an unverified OrgUser will not see admin features.

nickolashe:
@dannypernik did our recent convo help resolve your question above?

nickolashe:
#155

nickolashe:
#213

openCurrents/views.py
```diff
@@ -619,6 +619,33 @@ def get_context_data(self, **kwargs):
...
+        #check that user is part of org_admins group
+        userid = self.request.user.id
+        user = User.objects.get(id=userid)
+        org_admins_name = 'admin_' + str(orgid)
+        logger.info('org_admins_name: %s',org_admins_name)
+
+        try:
+            org_admins = Group.objects.get(name=org_admins_name)
+        except:
+            logger.info("Org exists without an org_admins group.")
```
nickolashe: lets log this at error level

openCurrents/views.py
```diff
@@ -2017,6 +2048,12 @@ def process_org_signup(request):
 
         try:
             org.save()
+
+            org_admins_name = 'admin_' + str(org.id)
```
nickolashe: I think the fact that this block of code is repeating the one above warrants factoring it out into a function

openCurrents/views.py
```diff
@@ -619,6 +619,33 @@ def get_context_data(self, **kwargs):
...
+
+        try:
+            org_admins = Group.objects.get(name=org_admins_name)
+        except:
+            logger.info("Org exists without an org_admins group.")
+            pass
+
+        if org_admins and not user.groups.filter(name=org_admins_name).exists():
+            logger.info("org admins group exists and user is not part of it.")
+            return redirect('openCurrents:500')
```
nickolashe: the error should be 403 (forbidden)

openCurrents/views.py
```diff
@@ -619,6 +619,33 @@ def get_context_data(self, **kwargs):
...
+        userid = self.request.user.id
+        user = User.objects.get(id=userid)
+        org_admins_name = 'admin_' + str(orgid)
+        logger.info('org_admins_name: %s',org_admins_name)
+
+        try:
+            org_admins = Group.objects.get(name=org_admins_name)
+        except:
+            logger.info("Org exists without an org_admins group.")
+            pass
```
nickolashe: this needs to return an http response of some kind, probably a 500 error page

openCurrents/views.py
```diff
@@ -619,6 +619,33 @@ def get_context_data(self, **kwargs):
 class AdminProfileView(LoginRequiredMixin, SessionContextView, TemplateView):
     template_name = 'admin-profile.html'
 
+    def get(self, request, *args, **kwargs):
+        context = super(AdminProfileView, self).get_context_data(**kwargs)
+        orgid = context['orgid']
```
nickolashe: i think we can improve on our current assumption of a single user to org mapping and check all such mappings for the purpose of authorizing the user

tsiology: @nickolashe can you expand on this? I think you are saying context['orgid'] should be a list of all orgid's the user/admin is associated with and we should check all org_admin groups for those ids.

tsiology:
@nickolashe I've removed the overridden get in AdminProfileView approach to instead set a context variable forbidden in get_context_data. This variable is then checked in the template. Functionality of redirecting users who are not in the group from admin-profile to 500 is working. Using 500 till 403 page is available.

openCurrents/templates/admin-profile.html
```diff
@@ -255,6 +254,14 @@ <h3 class="popup-title">Volunteers request approval</h3>
       });
     </script>
 
+
+    <script type="text/javascript">
```
nickolashe: @tsiology what if the user turns off javascript in their client?

nickolashe: i think we want to still stay within django realm

nickolashe: @tsiology see if you can override `render_to_response` method, which executes after `get_context_data` for class-based views:
https://docs.djangoproject.com/en/1.11/ref/class-based-views/mixins-simple/#django.views.generic.base.TemplateResponseMixin.render_to_response

nickolashe: also, FYI here is the list of class-based view methods listed in their execution order:
https://docs.djangoproject.com/en/1.11/ref/class-based-views/generic-display/#detailview

tsiology: would this override of render_to_response take the place of render in the get function as a way to pass all needed context to the template?

nickolashe: yes, i think it'll take place of the get. it returns the class instance of the view from what i understand. so no raw rendering like with render

nickolashe:
perhaps lets find an example of overriden render_to_response

openCurrents/views.py
```diff
@@ -88,9 +90,50 @@ def get_context_data(self, **kwargs):
         if userorgs:
             org = userorgs[0].org
             context['orgid'] = org.id
+            context['org_id'] = org.id
 
         return context
 
+class OrgAdminPermissionMixin(LoginRequiredMixin):
```
nickolashe: @tsiology @karbmk I put together this new mixin which can be used to assert admin permissions for any view

openCurrents/views.py
```diff
@@ -88,9 +90,50 @@ def get_context_data(self, **kwargs):
...
+    def dispatch(self, request, *args, **kwargs):
```
nickolashe: It overrides the default `dispatch` method for the view

PR #253, opened 2017-08-08
Defer hours

nickolashe:


nickolashe:
@karbmk @tsiology this a set of changes to the models needed to implement hour deferments by admins (as we discussed earlier). either of you please feel free to take on this from here.

karbmk:
Hi @nickolashe, was looking at the models for deffered times. Is user the admin and usertimelog the log which is not needed in the admin approve hours for that user?

nickolashe:
yes, exactly

openCurrents/views.py
```diff
@@ -187,6 +188,35 @@ def get_queryset(self):
             is_verified=False
         )
 
+        for g_d_t in get_defer_times:
+            if g_d_t.usertimelog in list(eventtimelogs):
```
nickolashe: let only form the list once, outside of the for loop

nickolashe: also, lets see if we can use a set since the lookup if much faster in that case

openCurrents/views.py
```diff
@@ -187,6 +188,35 @@ def get_queryset(self):
...
+                #list(eventtimelogs).remove(g_d_t.usertimelog)
+                eventtimelogs = eventtimelogs.filter(~Q(event=g_d_t.usertimelog.event))
```
nickolashe: @karbmk instead of iterating removing entries this way,why don't we form a list of deferred usertimelog id's and add an additional (excluding) filter when fetching eventtimelogs on line 181?

openCurrents/views.py
```diff
@@ -187,6 +188,35 @@ def get_queryset(self):
...
+        k = 0
+        #Check for usertimelogs ahead for upto a month
+        while k<5:
+            if not eventtimelogs:
+                week_startdate_monday = week_startdate_monday + timedelta(days=7)
+                today = timezone.now()
+
+                # build one weeks worth of timelogs starting from the oldest monday
+                time_log_week = OrderedDict()
+                get_defer_times = DeferredUserTime.objects.filter(user__id=userid)
```
nickolashe: looks like we duplicating lines 179-189 again here (can we turn this into a function)?

openCurrents/views.py
```diff
@@ -187,6 +188,35 @@ def get_queryset(self):
...
+                eventtimelogs = UserTimeLog.objects.filter(
+                    event__in=events
+                ).filter(
+                    datetime_start__lt=week_startdate_monday + timedelta(days=7) 
+                ).filter(
+                    datetime_start__gte=week_startdate_monday
+                ).filter(
+                    is_verified=False
+                )
+                for g_d_t in get_defer_times:
```
nickolashe: same comment as above

openCurrents/views.py
karbmk: Got you. I'm on it.

openCurrents/views.py
karbmk: I tried and exclude did'nt seem to work at all. But I will try again

openCurrents/views.py
karbmk: Yes, I think so. I will

karbmk:
Hi @nickolashe , taken care of the changes. Do review it.

openCurrents/views.py
nickolashe: @karbmk you mean you tried with an `.exclude` filter?

openCurrents/views.py
```diff
@@ -276,6 +249,37 @@ def get_queryset(self):
...
+            datetime_start__gte=week_startdate_monday
+        ).filter(
+            is_verified=False
+        )
+        exclude_usertimelog = []
+        for g_d_t in get_defer_times:
+            if g_d_t.usertimelog in eventtimelogs:
+                #eventtimelogs = eventtimelogs.filter(~Q(event=g_d_t.usertimelog.event))
+                exclude_usertimelog.append(g_d_t.usertimelog.event)
+        eventtimelogs = eventtimelogs.exclude(event__in=exclude_usertimelog)
```
nickolashe: @karbmk ok, i see you are using `.exclude` filter here now

nickolashe:
@karbmk I am now seeing pretty much the opposite behavior, where while an admin has hours to approve, there is no "Approve hours" button.

karbmk:
Hmm that's wierd. I will take a look on it!

openCurrents/views.py
karbmk: Yep. Also Approve hours button is displaying as expected when there are hours to be approved by the admin. Do you still see it working the opposite way??

nickolashe: @karbmk it is working now, but i needed to log out and log back in. not sure what caused this behavior.

karbmk: Will take a look for that. Not sure why that behaviour.

nickolashe:
@karbmk I think this looks good, so I am going to merge this. Lets open a separate PR to address #268.

PR #259, opened 2017-08-08
Approve hours status messages

tsiology:
Displays counts of volunteers approved/declined on subsequent approve-hours or admin-profile redirects.


openCurrents/templates/admin-profile.html
```diff
@@ -65,7 +85,6 @@ <h6 class="no-margin-bottom">My hours tracked: [ my_total ]</h6>
               <h6 class="one-margin-bottom">Events happening now</h6>
 
               {% for event in events_current %}
-                {% if event.event_type != "MN" %}
```
nickolashe: why are we getting rid of this check?

tsiology: event_current was updated in views to only be sending GR events, so this line became redundant.  We were printing the headers with no events listed below them because we were sending MN events from views, but just not showing them here.

nickolashe: @tsiology I see. Can we rename the context variable to `events_group_current` to reflect this?

tsiology: @nickolashe We can, but should then probably do the same for events_past and events_upcoming. It looks like events_upcoming has some additional applications in profile and admin profile. Thoughts?

nickolashe: @tsiology yep, I think so

tsiology: I've renamed the context variable in views and the admin-profile template to reflect this new GR filtering. Thanks.

openCurrents/templates/admin-profile.html
```diff
@@ -12,6 +12,26 @@
 
     <div id="container" class="js-close-any" canvas="container">
       <div id="wrapper">
+
+
+        {% if vols_approved or vols_declined %}
+          <div class="success row one-margin-top">
+            <div class="medium-8 columns small-centered">
+              {% if vols_approved|add:"0" >= 1 and vols_declined|add:"0" >= 1 %}
```
nickolashe: @tsiology what's the trick of adding a "0" ?

openCurrents/templates/approve-hours.html
```diff
@@ -11,119 +11,136 @@
...
+                               </a></p>
+
+                               <p class="volunteer-deferred hidden caution">Volunteer deferred. <a class="undo-defer">Undo</a></p>
+                             </div>
+                           </li>
+                         {% endfor %}
+                     {% endfor %}  
+
+
+                     <!--<li class="accordion-item is-active" data-accordion-item>
```
nickolashe: do we need to keep this code that's commented out?

openCurrents/views.py
```diff
@@ -164,8 +164,10 @@ def get_queryset(self):
         # week list holds dictionary ordered pairs for 7 days of timelogs
         week = []
 
-        # return nothing if unverified time logs not found
+        # return kwargs vols_approved and vols_declined if unverified time logs not found
         if not timelogs:
+            week = self.kwargs 
+            print(week)
```
nickolashe: @tsiology @karbmk lets only use logger when merging to develop

openCurrents/templates/admin-profile.html
tsiology: @nickolashe I found it was needed to make sure the variable is being compared as an integer.

openCurrents/templates/approve-hours.html
tsiology: talking with the others, it was a hard-coded example of MN tracked hours that is no longer needed. I've removed it.

openCurrents/views.py
nickolashe: @tsiology can you remove this then or switch it to use logger instead?

openCurrents/templates/approve-hours.html
```diff
@@ -206,6 +175,40 @@ <h3 class="popup-title"></h3>
...
       $(document).ready(function(){
+        if (document.getElementById("vols_approved_val") === null || document.getElementById("vols_declined_val") === null){
+
+        }
+        else{
+          var vols_approved = document.getElementById("vols_approved_val").value;
+          var vols_declined = document.getElementById("vols_declined_val").value;
+        
+
+          if(vols_approved >= 1 && vols_declined > 1)
```
nickolashe: @tsiology can we apply the `pluralize` tag to dry this up a bit?

openCurrents/urls.py
```diff
@@ -31,6 +31,7 @@
     url(r'^invite-friends/(?P<referrer>[\w\.@\+\-]*)$',
         views.InviteFriendsView.as_view(), name='invite-friends'),
     url(r'^approve-hours$', views.ApproveHoursView.as_view(), name='approve-hours'),
+    url(r'^approve-hours/(?P<vols_approved>\d+)/(?P<vols_declined>\d+)/$', views.ApproveHoursView.as_view(), name='approve-hours'),
```
nickolashe: @tsiology would it be better to pass these in the context instead of the url?

openCurrents/views.py
```diff
@@ -242,15 +245,19 @@ def get_queryset(self):
...
 
-    def post(self, request):
+    def post(self, request, **kwargs):
         """
         Takes request as input which is a comma separated string which is then split to form a list with data like
         ```['a@bc.com:1:7-20-2017','abc@gmail.com:0:7-22-2017',''...]```
         """
+        vols_approved = int(0)
+        vols_declined = int(0)
         post_data = self.request.POST['post-data']
```
nickolashe: i think our standard practice has been to handle this in the form instead of parsing the request dict directly here

openCurrents/views.py
```diff
@@ -331,9 +339,9 @@ def post(self, request):
             is_verified=False
         )
         if not timelogs:
-            return redirect('openCurrents:admin-profile')
+            return redirect('openCurrents:admin-profile', vols_approved, vols_declined)
                 
-        return redirect('openCurrents:approve-hours')
+        return redirect('openCurrents:approve-hours', vols_approved, vols_declined)
```
nickolashe: @tsiology @karbmk context can be provided to a view using django's `render` method

nickolashe:
@tsiology good work, just a couple of minor changes based on the comments above

openCurrents/views.py
```diff
@@ -1,4 +1,5 @@
 from django.contrib.auth import authenticate, login, logout
+            irint(week)
```
nickolashe: @tsiology I think this is a type

tsiology: Yes, not even sure how it got there but I see it as part of my last commit to change to logger. Will clean it

nickolashe: pushed the fix

openCurrents/templates/approve-hours.html
tsiology: @nickolashe I don't believe there is a javascript equivalent to pluralize, and I had to implement the messages here in javascript because the values for vols_approved/declined are accessed by iterating through the context from get_queryset which is sent as a list. Let me know if I need to explain the implementation more; I know it is now very straight-forward but found it was the best solution due to the complexity of data being passed from get_queryset.

tsiology: @nickolashe never mind, I see now it should work fine to use pluralize tag in javascript.

openCurrents/urls.py
tsiology: @nickolashe I may be going about this the wrong way, but I don't think using render to send vols_approved/declined as context instead of in the url with redirect is going to work. When I use render to show approve-hours again for a subsequent weeks, the template loads but is blank, as if get_queryset is failing to run from render. When the blank approve-hours template loads, it goes straight to admin-profile because there is no week_month timelog data.

nickolashe: @tsiology `render` needs context as an argument (which is simply a dictionary). see this for more info:
https://docs.djangoproject.com/en/1.11/topics/http/shortcuts/#example

tsiology: This is how I was calling render in the scenario above:

        context = {
            'vols_approved': vols_approved,
            'vols_declined': vols_declined
        }

return render(request, 'admin-profile.html', context)

tsiology: But instead of just vols_approved/declined, I guess what I need is for the render to acquire the same context as the get_queryset function in ApproveHoursView, which does not seem practical.

nickolashe: @tsiology i am not quite following, can you elaborate some more pls

tsiology: Sure, I will elaborate on my understanding. In ApproveHoursView, we have get_queryset that builds and returns our week of timelogs needing approval. week's data is parsed and evaluated in approve-hours template; if values from week are not found, we have no more timelogs needing approval and go to admin-profile. 

In the post function of ApproveHoursView, if more timelogs need approval, we need to go back to approve-hours and also pass the variables vols_approved/declined. This is currently being done as url args in redirect. I'm trying to use render with context data instead, but when I render approve-hours in post, I don't have all the week timelog data from get_queryset.

nickolashe: ok, i see. when we use `render` we are not actually using the view, but rendering the raw template. hence none of the view methods are relevant. one way around it that still avoids passing params in the url would be to post to some endpoint with these params as JSON, and that endpoint would then redirect to ApproveHoursView. but I am also fine to just punt on this and revert back to passing params in the url for now.

openCurrents/views.py
```diff
@@ -383,9 +390,9 @@ def post(self, request):
             is_verified=False
         )
         if not timelogs:
-            return redirect('openCurrents:admin-profile')
-                
-        return redirect('openCurrents:approve-hours')
+            return redirect('openCurrents:admin-profile', vols_approved, vols_declined)
+
+        return redirect('openCurrents:approve-hours', vols_approved, vols_declined)
```
nickolashe: what's the purpose of the above if statement if we redirecting to approve hours anyway?

openCurrents/urls.py
tsiology: Thanks for that explanation about rendering! I would like to see what we have make it to develop, but also understand the desire to move away from args in the url.

tsiology: Also still to handle Defer Volunteers.

openCurrents/views.py
tsiology: do you mean redirecting to admin-profile anyway?

nickolashe: ops, did not notice the distinction the first time, i see these are different

tsiology:
So then are we ok merging what is here for now?

nickolashe:
still have to test this locally

openCurrents/urls.py
nickolashe: @tsiology i don't understand your last comment. What needs those updates?

openCurrents/views.py
nickolashe: @tsiology can we handle this using forms?

nickolashe:
@tsiology Where are the counts supposed to be shown? I am not seeing any when I am redirected back to admin-profile.

nickolashe:
Furthermore, Approve hours button still stays active while I just confirmed all my choices. Do we have an outstanding issue for this? I made sure to merge develop branch as well.

tsiology:
Below is a screen on admin-profile with the status message after approving hours for one volunteer.

image

nickolashe:
@tsiology can you resolve the conflict please?

openCurrents/urls.py
tsiology: @nickolashe This PR adds messages for approved and declined hours, but does not yet handle deferred.

tsiology:
@nickolashe resolved conflicts in approve-hours template

openCurrents/views.py
nickolashe: bump

nickolashe: @karbmk @tsiology can you work together handling this using forms and JSON. Karthik jas already started working on this I believe. This link may be helpful:
https://stackoverflow.com/questions/14626702/django-forms-with-json-fields

nickolashe: so no manual splitting of posted data should be necessary

nickolashe:
@dannypernik @tsiology @karbmk is there anything holding us off on merging this?

dannypernik:
@nickolashe I've added the `{% forloop.first %}` here to display hours of first volunteer. This should be ready to merge.

tsiology:
I think this is ready.

PR #260, opened 2017-08-08
profile tweaks

tsiology:
app_hr profile url arg changed from True|False to 1|0.

issued_total and user_balance displays rounded to 2 decimal places.


openCurrents/urls.py
```diff
@@ -62,7 +62,7 @@
     url(r'^volunteering$', views.VolunteeringView.as_view(), name='volunteering'),
     url(r'^volunteer-requests$', views.VolunteerRequestsView.as_view(), name='volunteer-requests'),
     url(r'^profile$', views.ProfileView.as_view(), name='profile'),
-    url(r'^profile/(?P<app_hr>(True|False))/$', views.ProfileView.as_view(), name='profile'),
+    url(r'^profile/(?P<app_hr>(1|0))/$', views.ProfileView.as_view(), name='profile'),
```
nickolashe: @tsiology can you remind me what we are using this param for?

dannypernik: @nickolashe True if hours require approval & admin's first login of the week. True triggers approve-hours popup. @tsiology please confirm I've got that right

tsiology: Yes, that is the functionality I observed, though I believe @karbmk first created this. I swapped the setting and the checking of True|False for 1|0.

nickolashe: got it, thanks

PR #272, opened 2017-08-10
display GR events only

tsiology:
Fixing admin-profile display of MN events


nickolashe:
@tsiology I am getting an url resolution error on admin-profile for live-dashboard/<event_id>, because the event_id is null. Any idea why?

tsiology:
@nickolashe this is an error on the admin-profile page in this branch? I have not seen a similar problem

Issue #278, opened 2017-08-13
edit-event: Event time was changed and ended up a different time in the database.


dannypernik:
I noticed that Linda’s event was set for 4am to 6am. I received 2 emails the day before the event saying that the event details were changed. The first one said 2-4pm, and the next said 9am-11am. Neither of these match the event's listed time. Looks like it came from Linda’s account.

nickolashe:
i think i see the error, these times need to be explicitly assigned org's timezone:
https://github.com/opencurrents/opencurrents/blob/master/openCurrents/views.py#L848-L851

karbmk:
I'm on it

dannypernik:
Duplicate of #403

PR #284, opened 2017-08-14
Edit Event: Issue #278 fixed

karbmk:


openCurrents/views.py
```diff
@@ -799,12 +799,13 @@ def get_context_data(self, **kwargs):
         #get the event id from admin-profile page and fetch the data need for the UI
         context = super(EditEventView, self).get_context_data(**kwargs)
         # event
+        tz = OrgUser.objects.get(user__id=self.request.user.id).org.timezone
```
nickolashe: @karbmk this will fail for multiple user => org affiliations. Lets stick to our current way of handling this by picking the first such affiliation until we come up with a better way of handling this.

openCurrents/views.py
```diff
@@ -894,15 +895,16 @@ def post(self, request, **kwargs):
                         type(e)
                     )
                     return redirect('openCurrents:500')
+            tz = OrgUser.objects.get(user__id=self.request.user.id).org.timezone
```
nickolashe: same as above

openCurrents/views.py
karbmk: Got it

openCurrents/views.py
```diff
@@ -799,7 +799,11 @@ def get_context_data(self, **kwargs):
         #get the event id from admin-profile page and fetch the data need for the UI
         context = super(EditEventView, self).get_context_data(**kwargs)
         # event
-        tz = OrgUser.objects.get(user__id=self.request.user.id).org.timezone
+        org_user = OrgUser.objects.filter(user__id=self.request.user.id)
```
nickolashe: @karbmk lets factor this out into a standalone class, which will have methods like `get_user_org` and `get_user_org_timezone`. I think there are quite a few places where we are using this, so this will allow us to:
1. dry up the code quite a bit
2. whenever we are ready to support multiple org affiliations, we will just expand this module once without having to change the rest of the code

karbmk: Got it

openCurrents/views.py
```diff
@@ -517,6 +517,18 @@ class OfferView(TemplateView):
 class OrgHomeView(TemplateView):
     template_name = 'org-home.html'
 
+class OrgUserInformation():
```
nickolashe: @karbmk lets move this to its own file since this is not a view

karbmk: Sure. What should I keep the name of the file? ```orginfo.py``` or more general? Also, I'm assuming that i need to create the file in the same folder as views.py?

nickolashe: may be `orgs.py`? yes, same folder as views should work

karbmk: Got it. I'm on it

openCurrents/orgs.py
```diff
@@ -0,0 +1,12 @@
+class OrgUserInformation():
```
nickolashe: @karbmk should the class take `userid` at initialization?

openCurrents/views.py
```diff
@@ -15,6 +15,7 @@
 from datetime import datetime, time, date
 from collections import OrderedDict
 from copy import deepcopy
+from orgs import OrgUserInformation
```
nickolashe: @karbmk any other places in `views.py` where we can make use of this class?

karbmk: I think I have covered all of it. Atleast the ones which require orguser based on userid

openCurrents/orgs.py
karbmk: As of now no

openCurrents/views.py
```diff
@@ -108,9 +108,9 @@ def dispatch(self, request, *args, **kwargs):
         try:
             org_id = kwargs['org_id']
         except KeyError:
-            userorgs = OrgUser.objects.filter(user__id=user.id)
+            userorgs = OrgUserInfo().get_orguser(self.request.user.id)
```
nickolashe: @karbmk the whole point of using a class is that it does not have to be initialized each time. otherwise its the same as a function

openCurrents/orgs.py
```diff
@@ -1,12 +1,13 @@
-class OrgUserInformation():
-    def get_org_user(self, userid):
-        org_user = OrgUser.objects.filter(user__id=userid)
-        return org_user
+class OrgUserInfo(object):
```
nickolashe: @karbmk i've started making some changing for using the class properly, which the main reason why we wanted this. can you please follow apply the rest of the changes similarly.

nickolashe:
@karbmk lets double check all variable names here

nickolashe:
@karbmk have you tested the changes locally? i am getting a runtime error on views.py:102:
https://github.com/opencurrents/opencurrents/pull/284/files#diff-cdc3208e7bac1d05986424d1285c7fe1R102

because of the variable name change above

openCurrents/orgs.py
```diff
@@ -7,7 +7,7 @@ def __init__(self, userid):
         self.orgusers = OrgUser.objects.filter(user__id=userid)
 
     def get_orguser(self):
-        return self.orgusers[0] if self.orgusers else None
+        return self.orgusers if self.orgusers else None
```
nickolashe: @karbmk is this change consistent with how we use `OrgUserInfo` elsewhere?

karbmk: Yes i think so

PR #292, opened 2017-08-15
Login remember me

karbmk:
Implemented using sessions.
Files changed:
1. views.py

### DIFF
--- a/openCurrents/views.py
+++ b/openCurrents/views.py
@@ -89,6 +89,14 @@ class SessionContextView(View):
 
 class HomeView(SessionContextView, TemplateView):
     template_name = 'home.html'
+    def dispatch(self, *args, **kwargs):
+        try:
+            #If there is session set for profile
+            if self.request.session['profile']:
+                return redirect('openCurrents:profile')
+        except:
+            #If no session set
+            return super(HomeView, self).dispatch(*args, **kwargs)
 
 class NotFoundView(SessionContextView, TemplateView):
     template_name = '404.html'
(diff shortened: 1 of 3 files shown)


nickolashe:
@karbmk @dannypernik Are we looking to set some sort of timeout for this behavior?

karbmk:
No. As of now it has no timeout i think.

dannypernik:
@nickolashe there's another open issue regarding timeout

openCurrents/views.py
```diff
@@ -89,6 +89,14 @@ def get_context_data(self, **kwargs):
 
 class HomeView(SessionContextView, TemplateView):
     template_name = 'home.html'
+    def dispatch(self, *args, **kwargs):
+        try:
+            #If there is session set for profile
+            if self.request.session['profile']:
```
nickolashe: @karbmk i think login session is already stored in self.request.user. can't we just assert login by:
> if self.request.user.is_authenticated
?

nickolashe:
@karbmk good idea, please the question above

openCurrents/views.py
karbmk: But id the user has not clicked the keep me logged in button. Even then it would point to the profile which is not intended

dannypernik:
@nickolashe your question about timeout? Looks like he's replied

openCurrents/views.py
nickolashe: that's true, good point

PR #294, opened 2017-08-15
Event detail contact box

tsiology:
Updating event-detail contact box message depending of user is registered, org admin, or project coordinator. 

@dannypernik Danny to send new volunteer-messaged email template to Mandrill


tsiology:
@dannypernik I have the functionality of the registrants list pop up in place, but maybe you could help beautify the button? And then I think the last piece is making sure the Message button actually emails the registrants list if the user on the event-detail page is the coordinator.

dannypernik:
screen shot 2017-08-16 at 4 55 49 pm

@tsiology I registered 3 users to this event but I'm just getting this little symbol. We will want to display firstname lastname of each person. Also, we're going to need something like this to style the list of registrants: https://github.com/opencurrents/opencurrents/blob/master/openCurrents/templates/admin-profile.html#L81

tsiology:
Danny, for the empty list pictured above, can you verify you created UserEventRegistrations with is_confirmed = True for the event in question?

nickolashe:
bump

openCurrents/views.py
```diff
@@ -898,29 +898,58 @@ def _create_event(self, location, form_data):
             datetime_end=form_data['datetime_end'],
             coordinator_firstname=form_data['coordinator_firstname'],
             coordinator_email=form_data['coordinator_email'],
-            creator_id = self.userid
```
tsiology: @karbmk was the removal of creator_id intentional? It is used to notify admin of newly created events.

openCurrents/views.py
```diff
@@ -1356,11 +1479,50 @@ def get_context_data(self, **kwargs):
...
+        org_admin_group_name = '_'.join(['admin', str(context['event'].project.org.id)])
+
+        # group is supposed to exist at this point
+        try:
+            org_admin_group = Group.objects.get(name=org_admin_group_name)
+        except Group.DoesNotExist:
+            logger.error("org exists without an admin group")
+            return redirect('openCurrents:500')
+        if OrgUser.objects.filter(user__email = context['event'].coordinator_email).exists():
+            co_ordinator = False
```
tsiology: I think the logic here is reversed.

tsiology:
@dannypernik Is a list of emails acceptable for the registrants list? Also, I'm not sure what caused you to get the empty list you showed. Here is what I see currently, with a small tweak from my comment above.

image

openCurrents/views.py
tsiology: @nickolashe was the removal of creator_id intentional? It is used to notify admin of newly created events.

nickolashe: i don't think it was intentional.

nickolashe: @karbmk ?

openCurrents/views.py
nickolashe: i am having a hard time following the whole flow here. it'd be nice if there were more general as well as code-specific comments

nickolashe: perhaps just a confusing choice for the variable name since there is another boolean `is_coord` below

openCurrents/views.py
```diff
@@ -1269,23 +1326,27 @@ def get_context_data(self, **kwargs):
...
+        else:
+            co_ordinator = True
 
         is_admin = org_admin_group.user_set.filter(id=self.request.user.id).exists()
 
         # check if event coordinator
-        is_coord = Event.objects.filter(id=context['event'].id,coordinator_email=self.request.user).exists()
+        is_coord = Event.objects.filter(id=context['event'].id,coordinator_email=context['event'].coordinator_email).exists()
```
tsiology: This needs to remain checking if coordinator_email = self.request.user, right??

tsiology:
The event-detail contact box description and button are not updated for the following 6 cases:

1) non-registered non-admin volunteer  ==>   "Contact event coordinator"  "Register"
2)        registered non-admin volunteer  ==>   "Contact event coordinator"   "Send"
3) non-registered        admin  non-coord ==>  "Contact event coordinator - See registrants" "Register"
4)        registered        admin  non-coord ==> "Contact event coordinator - See registrants" "Send"   
5) non-registered  coordinator                ==>   "Contact registrants - See registrants" "Register"
6)        registered  coordinator                ==>   "Contact registrants - See registrants" "Send"

nickolashe:
@tsiology @karbmk I think 1-2 has already been taking care of in develop #336

nickolashe:
Lets combine 5 & 6 to "Contact registrants - See registrants" "Send"

tsiology:
I believe this meets requested changes and is now ready for review. Thanks!

openCurrents/views.py
```diff
@@ -1127,6 +1193,64 @@ def post(self, request, **kwargs):
...
+                            post_data['coordinator-email'],
+                            self.request.user.email
+                        )
+                    except Exception as e:
+                        logger.error(
+                            'unable to send transactional email: %s (%s)',
+                            e.message,
+                            type(e)
+                        )
+            except:
```
nickolashe: same comment as above

openCurrents/views.py
```diff
@@ -1404,11 +1528,48 @@ def get_context_data(self, **kwargs):
             event__id=context['event'].id,
             is_confirmed=True
         ).exists()
+
+        # check if admin for the event's org
```
nickolashe: @tsiology lets create a class (similar to OrgUserInfo) that will have a method `isOrgAdmin(userid)` and use it here

openCurrents/views.py
```diff
@@ -1404,11 +1528,48 @@ def get_context_data(self, **kwargs):
...
+ 
+        # list of confirmed registered users 
+        context['registrants'] = ''
+        if(is_coord or is_admin):
+            reg_list_uniques = []
+            reg_list_names = []
+            reg_list = UserEventRegistration.objects.filter(event__id=context['event'].id, is_confirmed=True)
+            
+            for reg in reg_list: 
+                if(reg.user.email not in reg_list_uniques):
```
nickolashe: i don't think we need this check any more because there is now a uniqueness constraint on that table

nickolashe: @tsiology on a side note, this is usually done using a `set()`. checking whether something is in the list in a loop is expensive because it has to traverse the entire list each time. so the cost is `O(n^2)` where n is the number of registrants

openCurrents/views.py
```diff
@@ -1592,6 +1753,42 @@ def event_register(request, pk):
     if form.is_valid():
         user = request.user
         message = form.cleaned_data['contact_message']
+
+        #check for existing registration
```
nickolashe: @tsiology what is the purpose of this change?

openCurrents/views.py
```diff
@@ -1592,6 +1753,42 @@ def event_register(request, pk):
...
+        is_registered = UserEventRegistration.objects.filter(user__id=user.id, event__id=event.id, is_confirmed=True).exists()
+        #check if the user is project coordinator
+        is_coord = Event.objects.filter(id=event.id,coordinator_email=user.email).exists()
+
+        #update is_confirmed=True or create new UserEventRegistration if needed
+        if not is_coord and not is_registered:
```
nickolashe: again, we don't need to check whether the volunteer is registered, we can just try registering while catching `IntegrityError` from the db

openCurrents/views.py
```diff
@@ -1618,46 +1823,48 @@ def event_register(request, pk):
                 'content': json.dumps(event.datetime_start,cls=DatetimeEncoder).replace('"','')
             }
         ]
-
-        user_event_registration = UserEventRegistration(
```
nickolashe: @tsiology please use this try statement, which is correct, to address the comment above

openCurrents/views.py
tsiology: @nickolashe do you mean the addition of registrants_names to context data? It is for displaying first and last name of registrants instead of their emails in the Registrant pop up on event-detail.

openCurrents/views.py
```diff
@@ -942,6 +942,64 @@ def _create_event(self, location, form_data):
             creator_id = self.userid
         )
         event.save()
+        try:
+            coord_user = User.objects.get(email=form_data['coordinator_email'])
+            if (coord_user.id != self.userid) and not OrgUser.objects.filter(user__id = coord_user.id).exists():
```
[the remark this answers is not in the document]
tsiology: @nickolashe I think this was actually @karbmk who implemented this check on OrgUser, and I'm wondering if we should be replacing the checks for OrgUser with checks for the user being in the org admin group?

openCurrents/views.py
tsiology: updated empty except blocks.

openCurrents/views.py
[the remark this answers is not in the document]
nickolashe: user being in org admin group would require our approval. I think what we need to check is whether the coordinator user has OrgUser affiliation with the given org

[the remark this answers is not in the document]
nickolashe: whatever the check we decide on, lets use the new OrgUserInfo class for that

dannypernik:
@tsiology apologies that I missed the comment about is_confirmed = True on UserEventRegistrations. What's the purpose of that checkbox again?

For the list styling, we want something like this:

`<ul>
  {% for registrant in registrants %}
    <li class="no-bullet">{{ registrant.firstname }} {{registrant.lastname }}</li>
  {% end %}
</ul>

<a class="registrants-box_close button round">Close</a>`

<li class="no-bullet"> puts the names in a vertical list and removes default bullet point. 

Also, @nickolashe @tsiology after I checked is_confirmed on the UserEventRegistrations, I tried logging in and got this error:

screen shot 2017-08-27 at 9 01 51 pm

I had previously run a python manage.py migrate and done something similar to this:

screen shot 2017-08-27 at 9 03 27 pm

This screenshot was taken after I tried to migrate after receiving the error above. However I think last time it was the deferredusertable that it said was stale.

nickolashe:
yeah i got the same error. this has to do with the db changes we are making for time-tracke_new. for some reason when going back to the current db state in develop it does not recreate deferredusertime table

nickolashe:
i think we need to follow these steps to fix this issue:
https://stackoverflow.com/questions/27583744/django-table-doesnt-exist

nickolashe:
@dannypernik @tsiology I think this is working for me now.

nickolashe:
@dannypernik did you see any problems testing this?

openCurrents/views.py
tsiology: @nickolashe would isOrgAdmin function fit inside OrgUserInfo, or can you expand on the new class you envision? I think we can get everything needed inside OrgUserInfo.

openCurrents/views.py
tsiology: no longer building registrants template variable as a unique list

openCurrents/views.py
tsiology: @nickolashe is there a strong preference to remove the check on is_registered and try building the UserEventRegistration regardless?

I am going to have to build the is_registered variable either way because it is needed to determine if the user should be sending a message to the coordinator as a new or existing registrant.

openCurrents/views.py
tsiology: I think isOrgAdmin function can be included in OrgUserInfo class, but must be passed the org.id.

tsiology:
@dannypernik updated registrants display as a vertical list with close button.

tsiology:
@nickolashe @dannypernik I think all requested changes are now in place for this PR.

openCurrents/views.py
nickolashe: no strong preference if you are building it anyway. otherwise the difference would be in an extra db call.

openCurrents/views.py
```diff
@@ -933,8 +933,9 @@ def _create_event(self, location, form_data):
         )
         event.save()
         try:
+            orguser = OrgUserInfo(self.userid)
             coord_user = User.objects.get(email=form_data['coordinator_email'])
-            if (coord_user.id != self.userid) and not OrgUser.objects.filter(user__id = coord_user.id).exists():
+            if (coord_user.id != self.userid) and not OrgUserInfo(coord_user.id).exists():
```
nickolashe: @tsiology what is this actually testing for?  OrgUserInfo is a class, I don't think it has a method `exists`. We may be "lucky" here and never see a runtime error because no specific exceptions are caught. I think you may have meant to say:
> if (coord_user.id != self.userid) and not OrgUserInfo(coord_user.id).get_orguser()

dannypernik:
@nickolashe @tsiology for some reason I can't start the server on this branch. I've gotten a couple different ImportErrors:

screen shot 2017-08-29 at 9 30 21 am

screen shot 2017-08-29 at 9 29 30 am

openCurrents/views.py
tsiology: Thanks, made this fix.

dannypernik:
Tim fixed the error I was seeing, I think we're good on this.

openCurrents/urls.py
```diff
@@ -101,7 +101,6 @@
         views.event_register_live, name='event_register_live'),
     url(r'^event_checkin/(?P<pk>\d+)/$', views.event_checkin, name='event_checkin'),
     url(r'^process_login/$', views.process_login, name='process_login'),
-    url(r'^org_user_list/$', views.org_user_list, name='org_user_list'),
```
nickolashe: @tsiology do you know what this was used for?

openCurrents/views.py
```diff
@@ -932,6 +932,65 @@ def _create_event(self, location, form_data):
...
+                            form_data['coordinator_email'],
+                            admin_user.email
+                        )
+                except Exception as e:
+                    logger.error(
+                        'unable to send transactional email: %s (%s)',
+                        e.message,
+                        type(e)
+                    )
+        except Exception as e:
```
nickolashe: @tsiology can we be more explicit with our expectations on the type of exception?

openCurrents/views.py
```diff
@@ -1121,6 +1188,64 @@ def post(self, request, **kwargs):
             elif post_data['event-privacy'] == '2':
                 edit_event.is_public = False
             edit_event.save()
+
+            try:
+                coord_user = User.objects.get(email=post_data['coordinator-email'])
+                if (coord_user.id != self.request.user.id) and not OrgUser.objects.filter(user__id = coord_user.id).exists():
```
nickolashe: looks like we can use OrgUserInfo here as well

openCurrents/views.py
```diff
@@ -1121,6 +1188,64 @@ def post(self, request, **kwargs):
...
+                            post_data['coordinator-email'],
+                            self.request.user.email
+                        )
+                    except Exception as e:
+                        logger.error(
+                            'unable to send transactional email: %s (%s)',
+                            e.message,
+                            type(e)
+                        )
+            except Exception as e:
```
nickolashe: same comment regarding the type of exception

openCurrents/views.py
```diff
@@ -1392,17 +1517,46 @@ def get_context_data(self, **kwargs):
...
             user__id=self.request.user.id,
             event__id=context['event'].id,
             is_confirmed=True
         ).exists()
+
+        # check if admin for the event's org
+        is_org_admin=orguser.is_org_admin(context['event'].project.org.id)
+
+        # check if event coordinator
+        is_coord = Event.objects.filter(id=context['event'].id,coordinator_email=self.request.user.email).exists()
```
nickolashe: @tsiology @karbmk can we add a method `is_event_coord` to orguser as well? i'd like us to abstract the low level logic where ever possible.

openCurrents/views.py
```diff
@@ -1392,17 +1517,46 @@ def get_context_data(self, **kwargs):
...
+
         context['is_registered'] = is_registered
+        context['admin'] = is_org_admin
+        context['coordinator'] = is_coord
+ 
+        # list of confirmed registered users 
+        context['registrants'] = ''
+        if(is_coord or is_org_admin):
```
nickolashe: > if is_coord or is_org_admin:

openCurrents/views.py
```diff
@@ -1392,17 +1517,46 @@ def get_context_data(self, **kwargs):
...
+
+            for email in reg_list:
+                reg_list_names.append( str(User.objects.get(email=email).first_name + " " + User.objects.get(email=email).last_name))
+ 
+            context['registrants_names'] = reg_list_names
 
         return context
 
 
+
```
nickolashe: remove

openCurrents/views.py
```diff
@@ -1586,6 +1740,42 @@ def event_register(request, pk):
     if form.is_valid():
         user = request.user
         message = form.cleaned_data['contact_message']
+
+        #check for existing registration
+        is_registered = UserEventRegistration.objects.filter(user__id=user.id, event__id=event.id, is_confirmed=True).exists()
```
nickolashe: @tsiology @karbmk or perhaps we can have another class called UserEvent, which would have methods like `is_registered` and `is_event_coord`? i think this would make more sense.

openCurrents/views.py
```diff
@@ -1586,6 +1740,42 @@ def event_register(request, pk):
...
+        #check if the user is project coordinator
+        is_coord = Event.objects.filter(id=event.id,coordinator_email=user.email).exists()
+
+        #update is_confirmed=True or create new UserEventRegistration if needed
+        if not is_coord and not is_registered:
+            try:
```
nickolashe: @tsiology i don't understand why we need a try / except block here. what can fail?

openCurrents/urls.py
tsiology: @nickolashe no, it was pulled in from some other branch by mistake in one of Danny's commits. Don't know from where.

openCurrents/views.py
```diff
@@ -1612,46 +1810,48 @@ def event_register(request, pk):
...
+                email_template = 'volunteer-messaged'
+                merge_var_list.append({'name': 'MESSAGE','content': message})
+                merge_var_list.append({'name': 'REGISTER','content': False})
+            elif not is_registered:
+                #message the coordinator as a new volunteer
+                email_template = 'volunteer-messaged'
+                merge_var_list.append({'name': 'MESSAGE','content': message})
+                merge_var_list.append({'name': 'REGISTER','content': True})
+        #if no message was entered and a new UserEventRegistration was created
+        elif(not is_registered and not is_coord):
```
nickolashe: > elif not is_registered and not is_coord:

openCurrents/views.py
tsiology: I hesitated on pulling in is_coord checking as a function in OrgUserInfo because it seemed less related to an OrgUser than Event.

openCurrents/views.py
tsiology: @karbmk Do you remember this try/except? The user_event_registration is built and saved regardless?

openCurrents/views.py
tsiology: I'm not sure what type of exception is expected, I'll need to do some study. Unless you or Karthik are familiar.

openCurrents/views.py
tsiology: done!

openCurrents/views.py
tsiology: This try except was meant to catch ObjectDoesNotExist exceptions for when a coordinator_email is given at create-event but that email does not exist as a user.  Updating the exception.

openCurrents/views.py
tsiology: updated

openCurrents/views.py
tsiology: Since the filter returns an empty list, not an exception if nothing is found, I'm removing this try/except.

tsiology:
@nickolashe I've updated the exceptions in question. I think maybe the EventInfo class discussion can be moved to a new issue so this PR can be merged?

nickolashe:
@tsiology sounds good, can you create that issue pls? i will go ahead and merge the PR.

PR #323, opened 2017-08-16
Modifications to invite vols from event and URL param change

karbmk:


nickolashe:
@karbmk what's the plan for switching to passing `JSON` in other places? do you think it makes sense to include this as part of this PR or another one?

karbmk:
I think we should do it in this PR itself. Once it's merged to develop then it would be easier that all JSON implementations are applied at once.

nickolashe:
@karbmk ok, so is this your plan then?

karbmk:
Yes it is!

nickolashe:
@karbmk I am removing "review requested" label since I think there are more places where we need to switch to passing `JSON`

karbmk:
Agreed

dannypernik:
@nickolashe @karbmk safe to delete branch?

PR #344, opened 2017-08-18
is_admin context var

nickolashe:


nickolashe:
@dannypernik @karbmk this will set a context var `is_admin` to True if and only if a user claimed employee or leader affiliation with an org and we have approved their claim.

openCurrents/views.py
```diff
@@ -98,7 +98,7 @@ def get_context_data(self, **kwargs):
 
         is_admin = False
         admin_org_group_names = [
-            '_'.join(['admin', str(userorg.id)])
+            '_'.join(['admin', str(userorg.org.id)])
```
nickolashe: @dannypernik this bug was the reason why admin access did not work for you

nickolashe:
@dannypernik i think you missed a few places where we need to use the new var:
https://github.com/opencurrents/opencurrents/pull/344/commits/ebbfaa3d50c7f3cfc400bd3a49077a94441621f8

nickolashe:
can you double check everything is working for you and merge if so

PR #346, opened 2017-08-18
Time tracker issue 327

karbmk:
Files changed:
1. views.py


nickolashe:
@karbmk good work!

dannypernik:
@nickolashe @karbmk can we delete this branch? How should we ensure we're deleting branches whenever it's safe and sensible to do so? Perhaps we can decide which branches we want to keep, then delete any others when they are merged?

nickolashe:
we should delete feature-specific branches upon merging

dannypernik:
@tsiology see last couple comments FYI

PR #350, opened 2017-08-19
Fix Skip on invite-volunteers

tsiology:
Including num_vols=0 in skip. Needed for msg on admin-profile


dannypernik:
@tsiology @nickolashe should branch be deleted?

nickolashe:
yes, as we with all feature-specific branches we should delete it

PR #353, opened 2017-08-22
Admin profile restyle

dannypernik:
Keep approve-hours button if no hours, just change styling. Also change some redirects to profile instead of admin-profile. @nickolashe looks like I may have accidentally included a migrations file here?


dannypernik:
Has been incorporated into design branch

PR #362, opened 2017-08-24
Track all admin actions on usertimelogs

nickolashe:


openCurrents/views.py
```diff
@@ -30,7 +30,7 @@
     Event, \
     UserEventRegistration, \
     UserTimeLog, \
-    DeferredUserTime
+    AdminActionUserTime
```
nickolashe: @karbmk @tsiology please take a look at the changes here. they mostly have to do with style / syntax, but there are also some comments regarding how i think we should refactor things. it'd be great if you can create issues corresponding to each of those while you are going through this.

PR #369, opened 2017-08-25
Time tracker new

karbmk:
**Time-tracker implementation:**
1. Must select an organisation. admin is an optional field if the volunteer doesn't know the admin.
2. if the org is selected and no admin is selected all the admins in that org are added as approvers in the ActionAdminUserTimelog.

**Approve-Hours implementation:**
1. Must have a record in ActionAdminUserTimelog with action_type = 'request_approval'.
2. Corresponding usertimelog from ActionAdminUserTimelog should have is_verified 'False'
3. declining the volunteer will update the ActionAdminUserTimelog of that usertimelog to 'dec' and doesn't delete the usertimelog record
4. Defer will update the ActionAdminUserTimelog of that usertimelog to 'def'
5. Approving the volunteer will update the ActionAdminUserTimelog of that usertimelog to 'app' and is_verified field of corresponding usertimelog to True.


openCurrents/forms.py
```diff
@@ -376,8 +391,12 @@ def clean(self):
         date_start = cleaned_data['date_start']
         time_start = cleaned_data['time_start']
         time_end = cleaned_data['time_end']
-        self.org = Org.objects.get(id=cleaned_data['org'])
-        tz = self.org.timezone
+        try:
```
nickolashe: @karbmk this is to handle the "Other org" choice in the form, correct?

openCurrents/forms.py
```diff
@@ -390,9 +409,9 @@ def clean(self):
             '%Y-%m-%d %I:%M%p'
         )
         cleaned_data['datetime_end'] = pytz.timezone(tz).localize(datetime_end)
-
-        if cleaned_data['datetime_start'] > cleaned_data['datetime_end']:
-            raise ValidationError(_('Start time needs to occur before End time'))
+        if datetime_end == datetime_start:
```
nickolashe: what's the reason why we are only testing for equality now?

openCurrents/templates/time-tracker.html
```diff
@@ -22,74 +22,61 @@
...
 
+              <div class="row one-margin-top">
+                <div class="small-6 columns">
+                  {{form.org}}
+                </div>
 
-          <div id="track-time" class="row hidden">
-            <div class="left medium-8 small-centered columns">
+                <div class="disabled small-6 columns">
+                  <!-- <select id="choose-org">
```
nickolashe: can we delete this?

nickolashe:
@karbmk can you provide a summary for the changes?

openCurrents/forms.py
karbmk: It was creating usertimelog for time difference 0 seconds so.

openCurrents/forms.py
karbmk: Yep that's right

openCurrents/templates/time-tracker.html
karbmk: Yeah we can. I had that commented just to make sure I append properly. We can delete it once the functionlity if working fine

openCurrents/forms.py
nickolashe: 0 seconds? is it rounding to seconds somewhere?

karbmk: No I mean like 12 am and 12 am should not be accepted but it was accepting so i put that equality constraint

nickolashe:
@karbmk as part of this PR, can we also make a change not to delete declined hours, but rather mark those with the 'declined' action type?

karbmk:
Yep that's happening. Sorry that I did'nt mention clearly in summary

openCurrents/views.py
```diff
@@ -372,23 +374,45 @@ def defer_month(self, week_startdate_monday, today):
...
+        # for g_d_t in get_defer_times:
+        #     if g_d_t.usertimelog in eventtimelogs:
+        #         #eventtimelogs = eventtimelogs.filter(~Q(event=g_d_t.usertimelog.event))
+        #         if g_d_t.user.id != userid or g_d_t.action_type != 'req':
+        #             exclude_usertimelog.append(g_d_t.usertimelog.id)
+        eventtimelogs = self.get_requested_vols(week_startdate_monday,events)#eventtimelogs.exclude(id__in=exclude_usertimelog)
+        return [eventtimelogs,time_log_week]
+
+    def get_requested_vols(self, week_date, events, user=None):
+        get_defer_times = AdminActionUserTime.objects.all()
```
nickolashe: lets use proper variable name, since these are not just deferred hours any more

karbmk: Sure

nickolashe: @karbmk this is going to get to be a huge object quickly. do we really need to fetch all records here?

nickolashe: why don't we just use filters right away here?

nickolashe: and why do we need both `get_defer_times` and `eventtimelogs`?

openCurrents/views.py
```diff
@@ -431,28 +457,41 @@ def post(self, request):
...
+                    # time_log = UserTimeLog.objects.filter(user=user
+                    #    ).filter(
+                    #       datetime_start__lt=week_date + timedelta(days=7)
+                    #    ).filter(
+                    #       datetime_start__gte=week_date
+                    #    ).filter(
+                    #       is_verified=False
+                    #    ).filter(
+                    #       event__in=events).delete()
+                    declined = self.get_requested_vols(week_date,events,user)
```
nickolashe: lets make the user specific call get_requested_vols really efficient. we don't need to fetch everyone's hours in that case (related to the comment above)

openCurrents/views.py
```diff
@@ -431,28 +457,41 @@ def post(self, request):
...
+                        #    ).filter(
+                        #       event__in=events
+                        # )
+                        # for g_d_t in get_defer_times:
+                        #     if g_d_t.usertimelog in eventtimelogs:
+                        #         #eventtimelogs = eventtimelogs.filter(~Q(event=g_d_t.usertimelog.event))
+                        #         if g_d_t.user.id != userid or g_d_t.action_type != 'req':
+                        #             exclude_usertimelog.append(g_d_t.usertimelog.id)
+                        approved = self.get_requested_vols(week_date,events,user)
+                        time_log = approved.update(is_verified=True)
```
nickolashe: @karbmk do you think we can switch to using just AdminActionUserTime of type `approved` for the purposes of asserting verified hours?

openCurrents/views.py
karbmk: Do you mean something like this:
``` AdminActionUserTime.objects.filter(
            usertimelog__event__in=events
        ).filter(
            usertimelog__datetime_start__lt=week_date + timedelta(days=7)
        ).filter(
            usertimelog__datetime_start__gte=week_date
        ).filter(
            usertimelog__is_verified=False
        ).filter(
            user__id=self.request.user.id
        ).filter(
            action_type='req'
        )```

File: openCurrents/views.py (hunk 6 above)
Comment (karbmk) [reply openCurrents/views.py]: Hmm Now I think so and applied filters instead of all the other for and usertimelog and it seems to work. I will update it and let you know again

File: openCurrents/views.py (hunk 5 above)
Comment (nickolashe) [reply openCurrents/views.py]: yep

**Comment by nickolashe** [comment]
@karbmk does the time tracker form submit the choice of "Someone else"? I am getting a runtime error on this line when submitting the form with the choice of "Someone else":
https://github.com/opencurrents/opencurrents/blob/time-tracker_new/openCurrents/views.py#L725

File: openCurrents/views.py (hunk 5 above)
Comment (nickolashe) [reply openCurrents/views.py]: the volunteer-specific one can further filter with `.filter(usertimelog__user_id=volunteer_user_id)

File: openCurrents/forms.py (hunk 2 above)
Comment (nickolashe) [reply openCurrents/forms.py]: would not the previous comparison take care of that since its a strict inequality?

**Comment by karbmk** [comment]
Oh that needs to be taken care off. I will work on that

File: openCurrents/views.py [hunk 7]
```diff
@@ -67,6 +67,13 @@ def diffInMinutes(t1, t2):
 def diffInHours(t1, t2):
     return round((t2 - t1).total_seconds() / 3600, 1)
 
+def org_user_list(request):
+    #returns the list of Org users for the provided org
+    input_org = request.POST['org']
```
nickolashe: @karbmk since this is not modifying the db, this should be a `GET` request with a return status code of 200.

openCurrents/views.py
```diff
@@ -246,11 +258,13 @@ def get_queryset(self):
...
-        ).filter(
-            is_verified=False
-        )
+        actiontimelog = AdminActionUserTime.objects.filter(user__id=userid).filter(action_type='req')
+        # timelogs = UserTimeLog.objects.filter(
+        #     event__in=events
+        # ).filter(
+        #     is_verified=False
+        # )
+        timelogs = actiontimelog.filter(usertimelog__event__in=events).filter(usertimelog__is_verified = False)
```
nickolashe: are we using `actiontimelog` anywhere else? why not combine all filters together to get `timelogs`?

openCurrents/views.py
```diff
@@ -260,8 +274,8 @@ def get_queryset(self):
             return week
 
         # find monday before oldest unverified time log
-        oldest_timelog = timelogs.order_by('datetime_start')[0]
-        week_startdate = oldest_timelog.datetime_start
+        oldest_timelog = timelogs.order_by('usertimelog__datetime_start')[0]
+        week_startdate = oldest_timelog.usertimelog.datetime_start
         week_startdate_monday = week_startdate - timedelta(days=week_startdate.weekday())
         today = timezone.now()
```
nickolashe: what does `self.defer_month` get us? does it really have to do anything with deferments?

nickolashe: a comment explaining the block of code below would be nice

openCurrents/views.py
```diff
@@ -290,25 +304,25 @@ def get_queryset(self):
             name = User.objects.get(username = user_email).first_name +" "+User.objects.get(username = user_email).last_name
 
             # check if same day and duration longer than 15 min
-            if timelog.datetime_start.date() == timelog.datetime_end.date() and timelog.datetime_end - timelog.datetime_start >= timedelta(minutes=15):
+            if timelog.usertimelog.datetime_start.date() == timelog.usertimelog.datetime_end.date() and timelog.usertimelog.datetime_end - timelog.usertimelog.datetime_start >= timedelta(minutes=15):
```
nickolashe: let `usertimelog = timelog.usertimelog` and use it in all if statements

openCurrents/views.py
```diff
@@ -355,23 +369,64 @@ def defer_month(self, week_startdate_monday, today):
...
         )
-        exclude_usertimelog = []
-        for g_d_t in get_defer_times:
-            if g_d_t.usertimelog in eventtimelogs:
-                #eventtimelogs = eventtimelogs.filter(~Q(event=g_d_t.usertimelog.event))
-                exclude_usertimelog.append(g_d_t.usertimelog.event)
-        eventtimelogs = eventtimelogs.exclude(event__in=exclude_usertimelog)
-        return [eventtimelogs,time_log_week]
+        if user:
+            get_requested_vol_times = get_requested_vol_times.filter(usertimelog__user=user)
```
nickolashe: lets use user_id rather than user

openCurrents/views.py
```diff
@@ -413,50 +469,82 @@ def post(self, request):
...
+                    # time_log = UserTimeLog.objects.filter(user=user
+                    #    ).filter(
+                    #       datetime_start__lt=week_date + timedelta(days=7)
+                    #    ).filter(
+                    #       datetime_start__gte=week_date
+                    #    ).filter(
+                    #       is_verified=False
+                    #    ).filter(
+                    #       event__in=events).delete()
+                    declined = self.get_requested_vols(week_date,events,user)
```
nickolashe: @karbmk looks like we are fetching usertimelogs in all three cases ('app', 'dec' and 'def'). can't we factor this step out of the if statement above?

openCurrents/views.py
```diff
@@ -502,18 +502,20 @@ def post(self, request):
...
-                                datetime_start__lt=week_date + timedelta(days=7)
-                            ).filter(
-                                datetime_start__gte=week_date
-                            ).filter(
-                                is_verified=False
-                            ).filter(
-                                adminactionusertime__user__id=self.request.user.id
-                            ).filter(
-                                adminactionusertime__action_type='app'
+                        usertimel = UserTimeLog.objects.prefetch_related(
```
nickolashe: proper indentation

openCurrents/views.py
```diff
@@ -971,19 +973,26 @@ def get_context_data(self, **kwargs):
         )
 
         # determine whether there are any unverified timelogs for admin
-        get_defer_times = AdminActionUserTime.objects.filter(user__id=userid)
-        exclude_usertimelog = []
-        timelogs = UserTimeLog.objects.filter(
+        usertimelogs = UserTimeLog.objects.filter(
```
nickolashe: @karbmk this checkes properly the existence of admin-specific requests so we can display the Approve hours button accordingly. we'll need to supplement this with non-admin specific requests at some point.

openCurrents/views.py
```diff
@@ -971,19 +973,26 @@ def get_context_data(self, **kwargs):
...
             event__in=events
+        ).annotate(
+            last_action_created=Max('adminactionusertime__date_created')
+        )
+
+        # admin-specific requests
+        admin_requested_hours = AdminActionUserTime.objects.filter(
```
nickolashe: @karbmk lets try to pick the order of filters so as to minimize the time of look up

nickolashe: notice that this time would usually be drastically different for different orderings of the filters

openCurrents/views.py
karbmk: @nickolashe , But what about approved. We have to modify even the usertimelogs in that case.

nickolashe:
@dannypernik @karbmk @tsiology The admin dropdown stopped working for me (upon selecting the org). Can you confirm you are seeing the same?

dannypernik:
@nickolashe the admin dropdown is working for me as long as there's an admin for the selected org.

nickolashe:
@karbmk can we resolve the conflicts please?

karbmk:
Done!

nickolashe:
@karbmk lets step through the following changes/fixes carefully together when we have a chance:
https://github.com/opencurrents/opencurrents/pull/369/commits/775e73bfa99beb9de8c7b2f052355b5a32f38456

openCurrents/templates/time-tracker.html
```diff
@@ -252,36 +252,47 @@ <h3 class="popup-title">Volunteer request</h3>
...
-              url: "{% url 'openCurrents:org_user_list' %}",
-              type:"POST",
-              async:false,
-              data : {
-                  csrfmiddlewaretoken: document.getElementsByName('csrfmiddlewaretoken')[0].value,
-                  org: strOrg
-              },
-              dataType : "json"
+              url: orgUserListUrl,
+              type: 'GET',
```
nickolashe: @karbmk as I mentioned before, any request that does not modify system state must be a `GET` request

openCurrents/templates/time-tracker.html
```diff
@@ -252,36 +252,47 @@ <h3 class="popup-title">Volunteer request</h3>
...
+          ).complete(function( res ) {
+              let data = res.responseJSON;
+
+              // populate select with org admin names
+              Object.keys(data).forEach(function (adminId) {
+                  let adminFirstName = data[adminId].firstname;
+                  let adminLastName = data[adminId].lastname;
+
+                  $('#id_admin_choice').append(
+                    `<option value=${adminId}>
```
nickolashe: @karbmk @tsiology Let's make sure that select options are always valued by id (or any other **unique** field)

openCurrents/views.py
```diff
@@ -252,77 +245,77 @@ class ApproveHoursView(OrgAdminPermissionMixin, SessionContextView, ListView):
...
-        # gather unverified time logs
-        #actiontimelog = AdminActionUserTime.objects.filter(user__id=userid).filter(action_type='req')
-        # timelogs = UserTimeLog.objects.filter(
-        #     event__in=events
-        # ).filter(
-        #     is_verified=False
-        # )
-        timelogs = AdminActionUserTime.objects.filter(
+        # fetch unverified time logs
+        requested_actions = AdminActionUserTime.objects.filter(
```
nickolashe: @karbmk lets use proper variable names so that:
1. another person can quickly grasp the code
2. the chance for bugs is reduced

nickolashe: in this case we are fetching adminaction records and not timelogs and this distinction caused several bugs in time-tracker and approve-hours

openCurrents/views.py
```diff
@@ -252,77 +245,77 @@ class ApproveHoursView(OrgAdminPermissionMixin, SessionContextView, ListView):
...
             user__id=userid
         ).filter(
             action_type='req'
         ).filter(
-            usertimelog__event__in=events
-        ).filter(
             usertimelog__is_verified = False
+        ).filter(
```
nickolashe: @karbmk currently we are fetching the entire list of manual events each time, so i moved the filter to be last

nickolashe: @karbmk can you create an issue to limit the events query by date pls

openCurrents/views.py
```diff
@@ -252,77 +245,77 @@ class ApproveHoursView(OrgAdminPermissionMixin, SessionContextView, ListView):
...
 
         # return kwargs vols_approved and vols_declined if unverified time logs not found
-        if not timelogs:
+        if not requested_actions:
             week = self.kwargs
-            #logger.info(week)
             return week
 
         # find monday before oldest unverified time log
-        oldest_timelog = timelogs.order_by('datetime_start')[0]
```
nickolashe: @karbmk lets try to avoid indexing query results directly

openCurrents/views.py
```diff
@@ -252,77 +245,77 @@ class ApproveHoursView(OrgAdminPermissionMixin, SessionContextView, ListView):
...
-                k=5
+                break
 
 
         time_log = OrderedDict()
         items = {'Total': 0}
 
-        for timelog in eventtimelogs:
-            user_email = timelog.user.email
-            name = User.objects.get(username = user_email).first_name +" "+User.objects.get(username = user_email).last_name
```
nickolashe: @karbmk there was a mistake here since its not actually a usertimelog anymore, but an adminaction record, we ended up presenting hours under the name of the admin, and not the volunteer

openCurrents/views.py
```diff
@@ -375,239 +368,166 @@ def get_queryset(self,**kwargs):
...
-        # return eventtimelogs.exclude(id__in=exclude_usertimelog)
-        get_requested_vol_times = AdminActionUserTime.objects.filter(
-            usertimelog__event__in=events
-        ).filter(
-            usertimelog__datetime_start__lt=week_date + timedelta(days=7)
+        requested_actions = self.get_requested_actions(week_startdate_monday, events)
+ 
+        return [requested_actions, time_log_week]
+
+    def get_requested_actions(self, week_date, events, user=None):
```
nickolashe: renamed the function for clarity

openCurrents/views.py
```diff
@@ -375,239 +368,166 @@ def get_queryset(self,**kwargs):
...
+        # fetches the volunteer-requested hours for admin review
+        logger.info(week_date)
+        requested_actions = AdminActionUserTime.objects.filter(
```
nickolashe: rearranged the filters for efficiency

openCurrents/views.py
```diff
@@ -375,239 +368,166 @@ def get_queryset(self,**kwargs):
...
-                    #        ).filter(
-                    #           is_verified=False
-                    #        ).filter(
-                    #           event__in=events)
-                    defered = self.get_requested_vols(week_date,events,user)
-                    #time_log = approved.update(is_verified=True)
-                    defered.update(action_type = 'def')
 
-        org = OrgUserInfo(userid)
+        admin_userid = self.request.user.id
```
nickolashe: since we have two different users in question for approving hours, lets be explicit about which user it is

openCurrents/views.py
```diff
@@ -375,239 +368,166 @@ def get_queryset(self,**kwargs):
...
+            i.split(':')[1] = '0' | '1'
+            i.split(':')[2] = '7-31-2017'
+            """
+            if not i:
+                continue
+
+            i = str(i)
+            # split the data for user, action_type, and date info
+            # action_type denotes approval or declining by admin
+            # TODO (@karbmk): since we are parsing raw request data,
```
nickolashe: @karbmk please note

openCurrents/views.py
karbmk: sure

openCurrents/views.py
karbmk: Yep got it.

openCurrents/views.py
```diff
@@ -375,239 +368,166 @@ def get_queryset(self,**kwargs):
...
+            # are we validating the input anywhere yet?
+            # Let's switch to doing this using forms ASAP
+            user = User.objects.get(username=i.split(':')[0])
+            action_code = int(i.split(':')[1])
+            action_type = action_type_map[action_code]
+            week_date = datetime.strptime(i.split(':')[2], '%m-%d-%Y')
+
+            # fetch volunteer requests for admin review
```
nickolashe: factored out this step outside of the if statement per my earlier comment

openCurrents/views.py
```diff
@@ -375,239 +368,166 @@ def get_queryset(self,**kwargs):
...
+   
+            if action_type == 'dec':
+                vols_declined += 1
+
+            # volunteer deferred
+            # TODO: decide if we need to keep this
+            elif action_type == 'def':
+                logger.warning('deferred timelog (legacy warning): %s', declined)
+
+            # TODO (@karbmk): instead of updating the requests for approval,
```
nickolashe: @karbmk can you create an issue for this please?

openCurrents/views.py
```diff
@@ -375,239 +368,166 @@ def get_queryset(self,**kwargs):
...
             is_verified=False
-        ).exclude(
-            deferments__id=userid
+        ).annotate(
+            last_action_created=Max('adminactionusertime__date_created')
         )
 
-        if not timelogs:
-            return redirect('openCurrents:admin-profile', vols_approved, vols_declined)
+        # admin-specific requests
```
nickolashe: since we can have several different adminaction records per usertimelog, we are looking for ones with type `request` that are the latest record for a given usertimelog

nickolashe: @karbmk can you create an issue to factor this out into a module to avoid repetition since we are using this code in AdminProfile for decorating "Approve hours" button as well

openCurrents/views.py
```diff
@@ -763,88 +683,97 @@ def track_hours(self, form_data):
...
+                    track_existing_datetime_start.astimezone(pytz.timezone(tz)).strftime('%-I:%M %p'),
+                    'to',
+                    track_existing_datetime_end.astimezone(pytz.timezone(tz)).strftime('%-I:%M %p'),
+                    'on',
+                    track_existing_datetime_start.strftime('%-m/%-d')
+                ])
+
+                return redirect('openCurrents:time-tracker', status_time)
+
+        if form_data['admin'].isdigit():
```
nickolashe: admin is passed by id now

openCurrents/views.py
```diff
@@ -763,88 +683,97 @@ def track_hours(self, form_data):
...
-                    self.isTimeLogValid = -1
-                    return
-        else:
-            self.isTimeLogValid = -1
-            return
+            return True
+
+        elif form_data['admin'] == 'other-admin':
+            # TODO (@karbmk): switch to using forms
+            admin_name = self.request.POST['admin_fname']
```
nickolashe: @karbmk lets transition to forms ASAP

openCurrents/views.py
karbmk: @nickolashe , doesn't this create a duplicate of the useraction except that we change the action_type?

openCurrents/views.py
```diff
@@ -763,88 +683,97 @@ def track_hours(self, form_data):
...
+                return False
+
+        elif form_data['admin'] == 'not-sure':
+            return False
 
-    def user_admin(self, orgid, usertimelog,admin):
-        #get and save the usertimelog for all admins in the ORG with orgid
-        #admin_user = OrgUser.objects.filter(org__id = orgid)
-        #for admin in admin_user:
+    def create_approval_request(self, orgid, usertimelog, admin_id):
```
nickolashe: renamed the function for clarity

openCurrents/views.py
```diff
@@ -763,88 +683,97 @@ def track_hours(self, form_data):
...
-        except:
-            user_new = User(
-                username=admin_email,
-                email=admin_email,
-                first_name=admin_name
-            )
-            user_new.save()
+            doInvite = not user_new.has_usable_password()
+        except User.DoesNotExist:
+            # user_new = User(
```
nickolashe: we decided not to track hours under unregistered admins yet

nickolashe: @karbmk @dannypernik are we letting the user know they should resubmit once the admin is registered?

openCurrents/views.py
```diff
@@ -957,26 +887,18 @@ def form_valid(self, form):
...
-                'to',
-                self.track_existing_datetime_end.astimezone(pytz.timezone(tz)).strftime('%-I:%M %p'),
-                'on',
-                self.track_existing_datetime_start.strftime('%-m/%-d')
-            ])
-            return redirect('openCurrents:time-tracker', status_time)
-
+            return redirect(
+                'openCurrents:time-tracker',
+                'You must know your admin. Sorry!'
```
nickolashe: @karbmk @dannypernik lets change the message to something more appropriate

openCurrents/views.py
```diff
@@ -1271,7 +1195,7 @@ def get_context_data(self, **kwargs):
         context = super(EditEventView, self).get_context_data(**kwargs)
         # event
         org_user = OrgUserInfo(self.request.user.id)
-        tz = org_user.get_org_timezone(self.request.user.id)
```
nickolashe: @karbmk @tsiology i am not sure why we were passing userid to `get_org_timezone`

openCurrents/views.py
```diff
@@ -2006,6 +1930,26 @@ def process_resend_verification(request, user_email):
     return redirect('openCurrents:check-email', user_email, status)
 
 
+@login_required
+def org_user_list(request, org_id):
```
nickolashe: refactored the endpoint into a GET request that returns a JSON dict keyed by admin id

nickolashe: @karbmk lets try to avoid any string manipulation on the output in favor of JSON dicts

openCurrents/views.py
karbmk: Yeah we shouldn't pass that!

openCurrents/views.py
karbmk: I was thinking that we create a user without password and send an invite. so that he could set the password and then he could find the hours directly in the approve-hours page once he logs in

openCurrents/views.py
karbmk: done

openCurrents/views.py
karbmk: got it

openCurrents/views.py
nickolashe: yes, by design we want to create a record for each admin action so we have full history. its not a duplicate by the definition of that model.

karbmk: Then I think we need to change the filtering I think. since we were filtering based on approval requests. We would have approval requests even though we have that approved. Any ideas on how that could be implemented?

karbmk: Oh I see that you are using ```annotate(
            last_action_created=Max('adminactionusertime__date_created')
        )```

nickolashe: @karbmk exactly, that makes sure if and only if the latest action for a given usertimelog is `req` it is presented for review to admin

PR #402, opened 2017-08-31
Handle form errors

nickolashe:


openCurrents/templates/time-tracker.html
```diff
@@ -28,6 +29,20 @@ <h3 class="title-sub">Record volunteer hours</h3>
         <form action ="" method="post">
           {% csrf_token %}
 
+          {% if form.errors %}
```
nickolashe: @dannypernik once we get this styled properly, it'd be nice if we can factor it out into a separate template and include in all forms

dannypernik: @nickolashe good call

dannypernik:
@nickolashe I've reverted changes attempting to use the red underline. We can add that back in later. I'd like to get the new error stylings merged today, then I think we're ready for Manual V1 launch!

PR #433, opened 2017-09-06
live-dashboard: fix checkin; admin-profile: report hours tracked by admin

nickolashe:


nickolashe:
reporting hours tracked by admin will work only for hours approved under the new manual tracking system and for any event hours starting from this change. we'll need to adjust previous hours manually for them to be included in the total.

nickolashe:
@dannypernik can you let me know if you are happy with the new hour totals pls

Issue #410, opened 2017-09-04
invite-volunteers & live-dashboard: email should not be case sensitive


dannypernik:
<img width="727" alt="screen shot 2017-09-04 at 11 31 01 am" src="https://user-images.githubusercontent.com/3347095/30034591-f4c1166a-9167-11e7-8460-3f63f319ded6.png">

Max and I ran several tests with same name, different name, and emails with and without capital letters. We concluded that we just need to make sure that emails with different capitalization are considered the same email address on both of these pages.

tevtonez:
here is the fix for invite-volunteers part: 008a9fd

MaxTattenbach:
@tevtonez Ok - I'll test once it has been merged to develop.  But yes I think that should solve the problem of multiple users being created.

tevtonez:
@MaxTattenbach thanks!

nickolashe:
@tevtonez i think the remaining step we discussed here is to add lower conversion to the js in live-dashboard

tevtonez:
@nickolashe yup, this fix is included in #964 PR

tevtonez:
I also checked time tracker form (as per discussion with Max) - looks like all fields are lowercased there too

PR #462, opened 2017-09-13
Issue #410 fix

karbmk:
added "iexact" in the query filters for ignoring the case.
built and tested for:
1. Signup
2. invite volunteers
3. Live-Dashboard


dannypernik:
This is looking good. Only thing that could be improved is that the status_msg that displays on admin_profile says that 1 volunteer has been invited even if the email already exists. I'm going to create a separate issue for this (It might be related to the inconsistent counts that @nickolashe was seeing previously - for example I know that when we tested emails with `< >` around them, it considered this text a separate email and added to the count.) I dont think we need to worry about this outstanding issue right now.

dannypernik:
Upon review, we already have an MVP issue for the status_msg, so I'll add my notes there.

nickolashe:
@karbmk would not it be easier to always convert emails to lower case in forms clean method?

karbmk:
You mean even while storing in the database you mean?

nickolashe:
yes, why don't we always store all emails lower case ?

Issue #11, opened 2017-04-05
Add opt-in on signup for receiving monthly updates


dannypernik:

nickolashe:
what's the status of this one?

dannypernik:
@nickolashe I thought this was working but I just tested and "Monthly updates" is not checked in my Account section of the db..

karbmk:
PR #470

PR #470, opened 2017-09-14
signup_confirm-account: issue #11 fixed

karbmk:
Data from HTML wasn't passed to the forms as expected which has been rectified.

### DIFF
--- a/openCurrents/forms.py
+++ b/openCurrents/forms.py
@@ -138,7 +138,6 @@ class EmailVerificationForm(forms.Form):
         user_password = cleaned_data.get('user_password')
         user_password_confirm = cleaned_data.get('user_password_confirm')
         validate_password_strength(str(user_password), str(user_password_confirm))
-        monthly_updates = cleaned_data['monthly_updates']
         # if user_password and user_password_confirm and user_password != user_password_confirm:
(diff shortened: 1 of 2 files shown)


openCurrents/templates/confirm-account.html
```diff
@@ -50,7 +50,7 @@ <h3 class="title-sub">Confirm account</h3>
           <div class="row center">
             <div class="small-12 columns">
               <div class="checkbox one-margin-bottom">
-                <input type="checkbox" name="" id="monthly-update" class="custom-checkbox" checked />
+                <input type="checkbox" name="monthly_updates" id="monthly-update" class="custom-checkbox" checked />
```
nickolashe: good find

openCurrents/forms.py
```diff
@@ -138,6 +138,7 @@ def clean(self):
         user_password = cleaned_data.get('user_password')
         user_password_confirm = cleaned_data.get('user_password_confirm')
         validate_password_strength(str(user_password), str(user_password_confirm))
+        monthly_updates = cleaned_data['monthly_updates']
```
nickolashe: @karbmk i don't we need this line since we are not modifying the variable?

karbmk: I will check on that.

PR #471, opened 2017-09-14
Edit event mvp issues 403, 374, 349 and 348

karbmk:
Yet to be completed


openCurrents/templates/edit-event.html
```diff
@@ -211,6 +211,7 @@ <h3 class="popup-title">Public or private?</h3>
 
         $('#location-box').append('<div id="location-'+id+'" class="relative hidden"><h6 class="fit-left navy">&</h6><span class="input-sizer"><input name="project-location-'+id+'" id="project-location-'+id+'" type="text" class="center" placeholder="location"/></span><i class="remove-input alert fa fa-lg fa-times"></i></div>');
         $('#location-'+id+'').slideDown();
+        initialize();
```
dannypernik: @karbmk does initialize() enable autocomplete for multiple locations? If so, I think we need it on create-event too. Thanks!

karbmk: Yes it does

openCurrents/templates/create-event.html
```diff
@@ -148,14 +148,13 @@ <h6 class="navy no-margin-bottom">Event coordinator</h6>
...
-                  {{ form.coordinator_firstname.errors }}
-                  {{ form.coordinator_firstname }}
-                </div>
-
-                <div class="input-right small-7 columns">
-                  {{ form.coordinator_email.errors }}
-                  {{ form.coordinator_email }}
+                <div class="input-left small-12 medium-8 columns end">
+                  <!-- form. .errors -->
+                  <select name="select-coordinator">
```
nickolashe: this does not look right

openCurrents/templates/edit-event.html
```diff
@@ -93,12 +104,14 @@ <h6 class="navy no-margin-bottom">Project coordinator</h6>
...
 
               <div class="row">
-                <div class="input-left small-5 columns">
-                  <input type="text" name="coordinator-name" placeholder="First name" value="{{event.coordinator_firstname}}"/>
-                </div>
-
-                <div class="input-right small-7 columns">
-                  <input type="text" name="coordinator-email" placeholder="Email" value="{{event.coordinator_email}}"/>
+                <div class="input-left small-12 medium-8 columns end">
+                  <!-- form. .errors -->
```
nickolashe: are we showing form errors anywhere?

openCurrents/views.py
```diff
@@ -1291,21 +1280,26 @@ def dispatch(self, request, *args, **kwargs):
...
+            data = form.cleaned_data
+        data = form.cleaned_data
         utc=pytz.UTC
         event_id = kwargs['event_id']
         edit_event = Event.objects.get(id=event_id)
 
         # save button - needs to handled through forms
         post_data = self.request.POST
-        if 'save-button' in post_data:
+        if 'save-button' in self.request.POST:
```
nickolashe: why not use form data?

openCurrents/views.py
```diff
@@ -1380,87 +1374,85 @@ def post(self, request, **kwargs):
...
-                        coord_email,
-                        self.request.user.email
-                    )
-                except Exception as e:
-                    logger.error(
-                        'unable to send transactional email: %s (%s)',
-                        e.message,
-                        type(e)
-                    )
+            # coord_user = None
```
nickolashe: why are we removing this?

nickolashe:
@karbmk why not simply use UpdateView here does most of the work for you (such as populating the form with the existing object's values)
https://docs.djangoproject.com/en/1.11/ref/class-based-views/generic-editing/#updateview

nickolashe:
@dannypernik I think the fact that you are combining update with a delete in the same template is making things more complicated for us

nickolashe:
@dannypernik @karbmk please familiarize yourself with the generic editing views in django:
https://docs.djangoproject.com/en/1.11/ref/class-based-views/generic-editing/#generic-editing-views

nickolashe:
this also https://rayed.com/wordpress/?p=1266

openCurrents/views.py
karbmk: Form data doesn't have save button

openCurrents/views.py
karbmk: That was because we eliminated the new admin signup from that page

openCurrents/templates/edit-event.html
karbmk: Nope. I should look on that

dannypernik: sorry I commented that out since I wasn't sure if it needed to change with the new input and it wouldn't run with django tags in the comments

dannypernik:
@nickolashe @karbmk  we can add a "Delete event" button on the dropdown of each event on profile page, though it's fairly common to see a "delete" button on edit pages. I'm happy with whichever is easier, we'd need to move the "Are you sure?" popup to the profile

karbmk:
Thanks for letting me know about the updateview @nickolashe . I will incorporate that. 

About the delete button we can call a different view during the "are you sure" pop-up. What do you guys think @nickolashe @dannypernik

openCurrents/templates/edit-event.html
nickolashe: i think this is pretty standard in other templates:
https://github.com/opencurrents/opencurrents/blob/develop/openCurrents/templates/time-tracker.html#L33-L45

openCurrents/views.py
nickolashe: is the purpose here that there are two forms and we want to tell which one was submitted?

openCurrents/views.py
nickolashe: are we still letting them to change the coordinator?

nickolashe:
@karbmk @dannypernik I think we should two separate views. In general, to maximize efficiency, I think we should go with whatever works well with the framework we've chosen to work with (django), irrespective of what's done on other sites.

dannypernik:
@nickolashe yes agreed. only thing I'm now confused about is that @karbmk says the "Are you sure" popup (from the edit-event page, presumably) can call a second view on that page? Also, the alternative that I'm imagining is having a "Delete event" button in the event dropdown on org-admin (formerly admin-profile). And then an "Are you sure" popup on org-admin.

Upon review, I prefer the second implementation. @nickolashe ?

openCurrents/templates/create-event.html
nickolashe: @karbmk what was the reason for this change and are there any plans on fixing this?

openCurrents/templates/edit-event.html
nickolashe: @karbmk can we include basic form error handling similar to the one linked above?

PR #505, opened 2017-09-20
offer / transaction models

nickolashe:


openCurrents/models.py
```diff
@@ -333,3 +296,64 @@ def __unicode__(self):
...
+
+
+class Offer(models.Model):
+    org = models.ForeignKey(Org)
+    item = models.ForeignKey(Item)
+    currents_share = models.DecimalField(
+        decimal_places=2,
+        max_digits=3
+    )
+    limit = models.IntegerField(default=-1)
```
nickolashe: need a separate table to keep number of transactions per vendor per month

PR #510, opened 2017-09-26
Issue#421: Add event coordinator & Issue#453: change registrants to invitees

dannypernik:


dannypernik:
@nickolashe We need to display first and last name of coordinator in place of [coordinator fullname] on event-detail

dannypernik:
@nickolashe I want to add a "Copy email list" button to the "See invitees" popup. Are the invitee emails currently accessible from the event-detail template? I'm thinking of passing in comma-separated emails into the copy-text with something like:

```
{% for email in invitees %}
{email},
{% end for %}
```
What variable names should I use for `email` and `invitees` (and is there somewhere in views.py or forms.py where I can find which variable names to use)?

nickolashe:
1. rename registrants to invitees on the frontend
2. close the pr

PR #523, opened 2017-10-12
Pending currents + copy email list

dannypernik:
Email list needs to be created with a forloop of all invitees/registrants' emails:
https://github.com/opencurrents/opencurrents/blob/pending_currents_emails/openCurrents/templates/event-detail.html#L108

Also, I know this is breaking the DRY principle (twice in one file), so perhaps you can show me how I can pass this into a function and use it for multiple copy/paste buttons:
https://github.com/opencurrents/opencurrents/blob/pending_currents_emails/openCurrents/static/js/copy.js


nickolashe:
sounds good, lets find some time to walk through these

nickolashe:
need to figure out copy/paste

PR #525, opened 2017-10-12
Transactions

dannypernik:
One question is how many offers we want displayed in marketplace before putting this live.

PR #528, opened 2017-10-14
Refactor business logic; Add unit tests

nickolashe:


openCurrents/ocuser.py
```diff
@@ -0,0 +1,82 @@
+from datetime import datetime
```
nickolashe: @rajatkapoor We'll first be unit testing this class, which now contains the business logic related to getting user balances and event data.

nickolashe: My plan is to add unit test scenarios and provide an example test tomorrow (10/14).

rajatkapoor: Ok.Sure

nickolashe:
merged into #525

PR #559, opened 2017-10-21
OcAuth

nickolashe:


openCurrents/views.py
```diff
@@ -557,7 +581,7 @@ def post(self, request, **kwargs):
             action_type='req'
         )        
 
-        redirect_url = 'approve-hours' if admin_requested_hours else 'admin-profile'
```
nickolashe: @dannypernik when you do a template url change, can you please do a project-wide substitution (as opposed to an one-off). found a bunch of active admin-profile references

nickolashe:
is_admin_(org|biz) vars
refactor authorization logic

openCurrents/views.py
dannypernik: Ah ok yes sorry about that

PR #582, opened 2017-10-22
Accounts / Transaction ledger

nickolashe:


rajatkapoor:
@nickolashe Had a look at the tests. I can start working on writing these. It would be great if you could tell me which modules to start with

dannypernik:
@nickolashe on the forms.py conflict here, I'm adding the `accept=image/*` line to only allow images to be selected on the file picker. This will prevent non-images from being selected by the user in the first place. However, I discovered that there are many more image filetypes accepted by image/* that we don't want to allow, so we should still use the error handling that currently checks image types.

nickolashe:
@dannypernik anything else holding off from merging?

openCurrents/forms.py
```diff
@@ -226,58 +257,80 @@ def __init__(self, *args, **kwargs):
...
-        })
-    )
-    date_start = forms.CharField(
+    event_date = forms.CharField(
         label='on',
-        widget=forms.TextInput(attrs={
-            'id': 'date_start',
-            'class': 'center'
+        widget=widgets.TextWidget(attrs={
+            'class': 'center',
```
dannypernik: TextWidget should already be passing in class=center

dannypernik: I removed center class from 3 fields in this branch and verified that TextWidget is passing it properly

openCurrents/templates/create-event.html
```diff
@@ -1,4 +1,4 @@
-{% load staticfiles %}
+event_date{% load staticfiles %}
```
dannypernik: I've removed this

openCurrents/forms.py
nickolashe: awesome, thanks

openCurrents/templates/org-admin.html
```diff
@@ -207,46 +200,17 @@ <h6 class="half-margin-bottom">Past events</h6>
...
-                  No hours have been reviewed. 
-                  <a href="{% url 'openCurrents:invite-volunteers' %}">
-                    Invite volunteers!
-                  </a>
-                </div>
-              </div>
-            {# {% endif %} #}
-          </div>
-        </div>
+        {% macro admin_hour_request_list_macro arg_list arg_status arg_timezone %}
```
nickolashe: @dannypernik does `layout` contain any refactoring using macros?

openCurrents/templates/admin-hour-request-list.html
```diff
@@ -21,7 +21,7 @@ <h6 class="half-margin-bottom">Hours {{ hours_status }}</h6>
 
                   {% if hours_status == 'requested' %}
                     <strong> {{ tl.user.first_name }} {{ tl.user.last_name }} </strong>
-                    recorded
+                    requested
```
nickolashe: @dannypernik I think you missed the fact that template is used both for requested and approved (passed as `hour_status` var)

openCurrents/templates/admin-hour-request-list.html
```diff
@@ -34,13 +34,12 @@ <h6 class="half-margin-bottom">Hours {{ hours_status }}</h6>
...
                 {% endwith %}
               </div>
-
             </div>
           {% endfor %}
         {% else %}
           <div class="row half-margin-bottom">
             <div class="left input-left small-9 columns">
-              No hours have been {{ hour_status }}.
+              No hours are pending approval.
```
nickolashe: @dannypernik I think you missed the fact this template is used both for requested and approved (through `hour_status` var)

dannypernik: Ok, when I checked this in org-admin, the Hours requested section said "No hours have been . Invite Volunteers!" So hour_status wasn't getting passed in.

PR #603, opened 2017-10-27
add event name to volunteer-messaged and coordinator-messaged emails

dannypernik:
@nickolashe please let me know when this is merged and I will update the email templates

### DIFF
--- a/openCurrents/views.py
+++ b/openCurrents/views.py
@@ -2026,6 +2026,10 @@ def event_register(request, pk):
             {
                 'name': 'EVENT_ID',
                 'content': event.id
+            },
+            {
+                'name': 'EVENT_NAME',
+                'content': event.project.name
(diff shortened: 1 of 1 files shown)


nickolashe:
this does not depend on anything else and can be merged right away?

dannypernik:
If you're asking if there's nothing else it depends on, no it doesn't. Or are you telling me that I can merge it to master without your approval? If I were to do so, am I correct in thinking I would then do a `git pull origin master` from develop branch to bring the changes there?

I will need to update the email templates to include this variable, but the variable needs to exist first. I got the variable passing into the email when I tested it in event_emails branch.

nickolashe:
@dannypernik yes that's correct. to sync develop with master you can just open a PR on gh after merging this to master.

PR #605, opened 2017-10-28
Redemption requests

nickolashe:


nickolashe:
@dannypernik i highly encourage you to review this pr

openCurrents/templates/biz-admin.html
```diff
@@ -107,102 +110,25 @@ <h6 class="half-margin-bottom">Our offers</h6>
...
-                  </div>
-
-                {% endwith %}
-              {% endfor %}
-
-            </div>
-          </div>
-        {% endif %}
-
-        {% if org_offers_redeemed %}
```
nickolashe: copy/pasting this section here was a direct violation of DRY principle again

openCurrents/templates/biz-admin.html
```diff
@@ -107,102 +110,25 @@ <h6 class="half-margin-bottom">Our offers</h6>
...
-                    </div>
-
-                    <div class="right relative small-3 columns">
-                      <a class="mini-open"><i class="blue half-margin-right fa-chevron-circle-down fa fa-lg"></i></a>
-                      <ul class="mini-menu">
-                        <a class="view-receipt-popup_open"><li>View receipt</li></a>
-                      </ul>
-                    </div>
-                  </div>
+        {% macro redemption_list_macro arg_list arg_status %}
```
nickolashe: i factored out the block into another template called `redemption-list.html` and i am overriding its `redemption_list` block twice here

nickolashe: because django does not allow for using the same block name more than once in the same template, i had to use a package called `django-templates-macros`

nickolashe: one can define a macro with params that are passed into the block

nickolashe: this passes django's check since the macro is only defined once

openCurrents/templates/biz-admin.html
```diff
@@ -107,102 +110,25 @@ <h6 class="half-margin-bottom">Our offers</h6>
             {% else %}
                 <div class="row">
                   <div class="left small-12 columns">
-                    <p>[ org_name ] has not made any offers. <a href="{% url 'openCurrents:offer' %}">Create one!</a></p>
+                    <p>{{ orgname }} has not made any offers. <a href="{% url 'openCurrents:offer' %}">Create one!</a></p>
```
nickolashe: `orgname` is available in every template, why are we not using it?

nickolashe:
i think the kind of template inheritance i used here can be applied to other templates

openCurrents/templates/profile.html
```diff
@@ -115,32 +115,37 @@ <h6 class="half-margin-bottom">Status</h6>
 
             {% if offers_redeemed %}
               {% for offer_redeemed in offers_redeemed %}
-                {% with offer=offer_redeemed.transaction.offer %}
+                {% with transaction=offer_redeemed.transaction status=offer_redeemed.get_action_type_display %}
                   <div class="row half-margin-bottom">
                     <div class="left small-9 columns">
-                      (10/15) - 
+                      {{ offer_redeemed.date_updated|day:timezone }} -
```
nickolashe: i forgot that we already have a tag for proper display of dates

openCurrents/templates/profile.html
```diff
@@ -115,32 +115,37 @@ <h6 class="half-margin-bottom">Status</h6>
...
-                      {% if offer_redeemed.get_action_type_display == 'pending' %}
-                        offers ${# disc|percent_price:0.01 #} for 
-                        <span class="no-wrap">
-                          <img class="h6-symbol" src="{% static 'img/symbol-navy.svg' %}"/>{# disc|percent_price:0.001 #}
-                        </span>
-                      {% else %}
-                        accepted <img class="h6-symbol" src="{% static 'img/symbol-navy.svg' %}"/>{# disc|percent_price:0.001 #}
-                        for ${# disc|percent_price:0.01 #}
-                      {% endif %}
+                      {% with price=transaction.price_reported percent_share=transaction.offer.currents_share|percent_to_price %}
```
nickolashe: new `percent_to_price` tag converts currents share to percent

openCurrents/templates/profile.html
```diff
@@ -115,32 +115,37 @@ <h6 class="half-margin-bottom">Status</h6>
...
+                        {% with price_share=price|mult:percent_share %}
```
nickolashe: `mult` tag just multiplies two numbers

openCurrents/templates/profile.html
```diff
@@ -115,32 +115,37 @@ <h6 class="half-margin-bottom">Status</h6>
...
+                          {% if status == 'pending' %}
+                            offers ${{ price_share|floatformat:2 }} for
```
nickolashe: we can use django's builtin `floatformat` tag to pick the number of decimals to display

openCurrents/templates/profile.html
```diff
@@ -115,32 +115,37 @@ <h6 class="half-margin-bottom">Status</h6>
...
+                            <span class="no-wrap">
+                              <img class="h6-symbol" src="{% static 'img/symbol-navy.svg' %}"/>{{ price_share|usd_to_current|floatformat:3 }}
```
nickolashe: the new `usd_to_current` does the conversion, which we then display with 3 decimals

openCurrents/templates/biz-admin.html
dannypernik: At the time of design I didn't know that

openCurrents/templates/biz-admin.html
dannypernik: Yes. It seems like we've been in agreement that we need to complete all MVP issues before refactoring code. So this is temporary until I'm able to refactor all templates at once. Is that ok?

openCurrents/templates/profile.html
dannypernik: ok so this does not change the variable value, just the display, correct? Are we therefore saving Currents to indefinite digits?

openCurrents/templates/biz-admin.html
nickolashe: i am not sure its ok to add new undry code at this point. please take a careful look at the changes in this template to see how i approached the problem from a dry perspective.

nickolashe: click show outdated diffs in biz-admin.html if you have not seen those yet

nickolashe: or step through the commit history one by one

dannypernik: @nickolashe ok I'll start using blocks for outstanding issues. Not clear yet on these tags `{% usemacro redemption_list_macro redeemed_pending 'Pending' org_timezone %}` and also receiving error on biz-admin:
image

openCurrents/templates/profile.html
nickolashe: just the display

nickolashe: we can pick the precision we want in the db

PR #614, opened 2017-10-31
sync develop with changes pushed directly to master

dannypernik:


dannypernik:
@nickolashe circleci is giving a pip install error, not sure the cause?

PR #627, opened 2017-11-02
Use forms.py on edit-event, move delete-event-popup to org-admin

dannypernik:
Apparently my attempts to disable the 'next' buttons in redeem-currents were contained in this branch as well - it didn't work out but some things got moved around there.


nickolashe:
@dannypernik this gets the form fields displayed properly. hoping you can study this example and then style appropriately.

nickolashe:
we also need to change CreateEventForm similarly since I changed coordinator to be a pointer to a user object instead of firstname / email as before.

openCurrents/forms.py
```diff
@@ -310,6 +312,106 @@ def clean(self):
         return cleaned_data
 
 
+class EditEventForm(forms.Form):
+    def __init__(self, *args, **kwargs):
+        event_id = kwargs.pop('event_id')
```
nickolashe: this a way to pass event id to the form so its available for internal use inside form

openCurrents/forms.py
```diff
@@ -310,6 +312,106 @@ def clean(self):
...
+        self.event = Event.objects.get(id=event_id)
+        self.org = self.event.project.org
+        tz = self.org.timezone
+
+        # call parent init in order to be able to access form fields
+        super(EditEventForm, self).__init__(*args, **kwargs)
+
+        # populate initial values for event
```
nickolashe: prepopulate the initial values based on fetched event object

openCurrents/forms.py
```diff
@@ -310,6 +312,106 @@ def clean(self):
...
+        self.fields['event_date'].initial = self.event.datetime_start.astimezone(
+            pytz.timezone(tz)
+        ).date()
+        self.fields['event_starttime'].initial = self.event.datetime_start.astimezone(
+        pytz.timezone(tz)
+        ).time()
+        self.fields['event_endtime'].initial = self.event.datetime_end.astimezone(
+        pytz.timezone(tz)
+        ).time()
+        self.fields['event_privacy'].initial = int(self.event.is_public)
```
nickolashe: is_public is a boolean so it needs to be converted to an int (see the def for privacy choices below)

openCurrents/forms.py
```diff
@@ -537,7 +639,7 @@ def __init__(self, *args, **kwargs):
         offer_id = kwargs.pop('offer_id')
         self.offer_init = Offer.objects.get(id=offer_id)
 
-        super(OfferEditForm, self).__init__(*args, **kwargs)
```
nickolashe: @dannypernik have you been able to set up a linter in sublime editor?

openCurrents/migrations/0061_auto_20171103_0741.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
nickolashe: migration file produced by makemigrations to use a user object for coordinator

openCurrents/templates/edit-event.html
```diff
@@ -33,14 +33,18 @@ <h3 class="title">Edit event</h3>
             <div class="small-12 medium-6 small-centered columns">
 
               <h6 class="fit-left navy half-margin-top">Let's...</h6>
-              <span class="input-sizer"><input name="project-name" class="center" type="text" placeholder="do some good" value="{{ event.project.name }}"/></span>
+              <span class="input-sizer">
```
nickolashe: does this have to be a span or can we move the class into the input?

openCurrents/templates/edit-event.html
```diff
@@ -63,75 +71,56 @@ <h6 class="fit-left navy half-margin-top">to</h6>
...
-              <span class="input-sizer"><input name="project-location-1" id="project-location-1" type="text" class="center" placeholder="location" value="{{event.location}}"/></span>
+              <span class="input-sizer">
+                {{ form.event_location }}
+                <!-- <input name="event-location" id="event-location-1" type="text" class="center" placeholder="location" value="{{event.location}}"/> -->
+              </span>
             </div>
           </div>
 
-          <a id="add-location"><i class="fa fa-plus-circle half-margin-right"></i>Add another location</a>
+          <!-- <a id="add-location"><i class="fa fa-plus-circle half-margin-right"></i>Add another location</a> -->
```
nickolashe: add location is not needed on edit-event, i was surprised to see this here still

openCurrents/templates/edit-event.html
```diff
@@ -63,75 +71,56 @@ <h6 class="fit-left navy half-margin-top">to</h6>
...
 
           <div class="row">
             <div class="small-6 small-centered columns">
               <input type='submit' name='save-button' class="one-margin-top button round" value="Save event">
-              <p><a class="button round tiny alert">Delete event</a></p>
              <!--  <p><a class="template-popup_open">Save template...</a></p> -->
             </div>
           </div>
-
-          <div class="delete-event-popup" class="modal center small-12 medium-9 large-7 small-centered columns">
```
nickolashe: @dannypernik lets try to never combine two forms together for two different actions in the same template (edit and delete in this case)

openCurrents/templates/org-admin.html
```diff
@@ -144,7 +142,7 @@ <h6 class="half-margin-bottom">Upcoming events</h6>
                       <a href="{% url 'openCurrents:event-detail' event.id %}"><li>View & contact invitees</li></a>
                       {# {% endif %} #}
                       <a href="{% url 'openCurrents:invite-volunteers' event.id %}"><li>Invite volunteers</li></a>
-                      <!-- <a href="{% url 'openCurrents:edit-event' event.id %}"><li>Edit event</li></a> -->
+                      <a href="{% url 'openCurrents:edit-event' event.id %}"><li>Edit event</li></a>
```
nickolashe: added link back in

openCurrents/templates/profile.html
```diff
@@ -302,7 +302,8 @@ <h3 class="popup-title">No dollars available</h3>
         <h3 class="popup-title">Confirm cash out</h3>
```
nickolashe: why is this in this pr?

openCurrents/templates/redeem-currents.html
```diff
@@ -95,6 +95,7 @@ <h4 class="one-margin-bottom">Upload receipt or proof of purchase</h4>
                 </p>
```
nickolashe: @dannypernik can you move this to a separate pr?

openCurrents/views.py
```diff
@@ -1395,207 +1396,206 @@ def get_form_kwargs(self):
 
 
 # needs to be implemented using UpdateView
-class EditEventView(OrgAdminPermissionMixin, SessionContextView, TemplateView):
+class EditEventView(OrgAdminPermissionMixin, SessionContextView, FormView):
```
nickolashe: @dannypernik the reason the form was not accessible in the template for you is because this was a `TemplateView`, which means it needs to be passed explicitly. Usually we use `FormView` with forms, in which case it is passed by default. `FormView` needs to implement `form_valid` method which handles valid form data (see below)

openCurrents/views.py
```diff
@@ -1395,207 +1396,206 @@ def get_form_kwargs(self):
...
+                    type(e)
+                )
+
+        project = Project.objects.get(id = edit_event.project.id)
+        project.name = str(post_data['project-name'])
+        project.save()
 
         return redirect('openCurrents:org-admin')
 
+    def get_form_kwargs(self):
```
nickolashe: this method is important when you need to pass some object id's. in this event id is passed and event is retrieved in the form __init__ method (see above)

openCurrents/views.py
```diff
@@ -1395,207 +1396,206 @@ def get_form_kwargs(self):
...
-        event = Event.objects.get(id=event_id)
-        context['event'] = event
-        context['start_time'] = str(event.datetime_start.astimezone(pytz.timezone(tz)).time())
-        context['end_time'] = str(event.datetime_end.astimezone(pytz.timezone(tz)).time())
-        context['date_start'] = str(event.datetime_start.astimezone(pytz.timezone(tz)).date())
+
         return context
 
     def dispatch(self, request, *args, **kwargs):
         event_id = kwargs.get('event_id')
```
nickolashe: kwargs dictionary in any view's `dispatch` method contains the url params for that view (event_id in this case)

openCurrents/views.py
```diff
@@ -1395,207 +1396,206 @@ def get_form_kwargs(self):
...
-        if timezone.now() > event.datetime_end:
-            return redirect('openCurrents:org-admin')
+        self.event = Event.objects.get(id=event_id)
+        if timezone.now() > self.event.datetime_end:
+            return redirect('openCurrents:403')
         else:
             return super(EditEventView, self).dispatch(request, *args, **kwargs)
 
-    def post(self, request, **kwargs):
+    def form_valid(self, form):
```
nickolashe: this still needs work

openCurrents/forms.py
dannypernik: Is `self` in python kind of like `this` in JS?

openCurrents/forms.py
dannypernik: I have not, I believe the ones I looked into were making more changes than we were looking for. I'll look again though. Are we looking for any other features from a linter than correct spacing?

openCurrents/migrations/0061_auto_20171103_0741.py
dannypernik: Is this the line that uses a user object for coordinator?
https://github.com/opencurrents/opencurrents/pull/627/files/672087927b795b8f37feb44e8521b2d4f4de2efb#diff-7c6a6fca15691185cd037898bd86f401R123

openCurrents/templates/edit-event.html
dannypernik: needs to wrap the input

openCurrents/templates/edit-event.html
dannypernik: Right, will move it to the mini-menu on org-admin

openCurrents/templates/profile.html
```diff
@@ -254,8 +254,8 @@ <h4 class="popup-title">Welcome to the openCurrents beta</h4>
             <p>Are you willing to help us connect with businesses where you shop, nonprofits where you volunteer, or friends who might be interested?</p>
 
             <p class="three-halves-margin-top">
-              <a type="submit" class="button round">Yes</a>
-              <a type="submit" class="button round secondary">No</a>
+              <button type="submit" class="button round">Yes</button>
+              <button type="submit" class="button round secondary">No</button>
```
dannypernik: I don't know why this is showing as a diff because I've already changed `a` to `button` in develop

openCurrents/forms.py
nickolashe: yes, exactly the same (it refers to the class itself)

nickolashe: so `self.fields` refers to the fields of an instance of `EditEventForm`

openCurrents/forms.py
nickolashe: they should be configurable. we can come up with a basic set of syntax rules.

nickolashe: here are some if you'd like to configure:
1. no extra spaces
2. lines no longer than 80 chars
3. double new line between class definitions

openCurrents/migrations/0061_auto_20171103_0741.py
nickolashe: yes

openCurrents/templates/edit-event.html
nickolashe: what does it do?

openCurrents/templates/profile.html
nickolashe: can you provide a link from develop in support of this pls

nickolashe:
@dannypernik i refactored create and edit event forms in a more intuitive way and also provided extensive comments

openCurrents/forms.py
```diff
@@ -312,17 +351,22 @@ def clean(self):
         return cleaned_data
 
 
-class EditEventForm(forms.Form):
+class EditEventForm(CreateEventForm):
```
nickolashe: note that now `EditEventForm` inherits from `CreateEventForm`

openCurrents/forms.py
```diff
@@ -337,79 +381,27 @@ def __init__(self, *args, **kwargs):
         self.fields['event_location'].initial = self.event.location
         self.fields['event_description'].initial = self.event.description
 
-        # obtain the list of approved org admins to populate
```
nickolashe: factored out into a function inside `CreateEventForm`

openCurrents/templates/profile.html
dannypernik: https://github.com/opencurrents/opencurrents/blob/develop/openCurrents/templates/profile.html#L257

openCurrents/forms.py
```diff
@@ -228,58 +245,80 @@ def __init__(self, *args, **kwargs):
...
-    coordinator_firstname = forms.CharField(
-        widget=forms.TextInput(attrs={'placeholder': 'First name','value':''})
+    event_privacy = forms.ChoiceField(
+        widget=forms.RadioSelect(
+            attrs={
+                'class': 'custom-radio',
+                'id': 'id-event-privacy'
+            }
+        ),
+        choices=[(1, 'public'), (0, 'private')],
```
dannypernik: noticed that 1 is public here, though it looks like the inputs may have been reversed in the HTML template:
https://github.com/opencurrents/opencurrents/blob/edit_event_form/openCurrents/templates/create-event.html#L123

openCurrents/templates/create-event.html
```diff
@@ -82,23 +82,23 @@ <h6 class="fit-left navy half-margin-top"> {{ form.date_start.label }} </h6>
             <div class="small-12 medium-6 small-centered columns">
               <div class="row">
                 <div class="input-left small-6 columns">
-                  {{ form.time_start.errors }}
-                  <h6 class="fit-left navy half-margin-top"> {{ form.time_start.label }} </h6>
-                  <span class="input-sizer"> {{ form.time_start }} </span>
-                  {% if form.time_start.help_text %}
+                  {{ form.event_starttime.errors }}
+                  <h6 class="fit-left navy half-margin-top"> {{ form.event_starttime.label }} </h6>
```
dannypernik: Are you still onboard with leaving labels as text in HTML? Even if a Django label automatically defines a label correctly, "Start time" is shorter than {{ form.event_starttime.label }}.

openCurrents/templates/create-event.html
```diff
@@ -120,25 +120,25 @@ <h6 class="fit-left navy half-margin-top"> at </h6>
           <a id="add-location"><i class="fa fa-plus-circle half-margin-right"></i>Add another location</a>
 
           <div class="row three-halves-margin-top">
-              {{form.is_public.0.tag}}
-              <label for="id_is_public_0">Public event</label>
+              {{form.event_privacy.0.tag}}
```
dannypernik: From above comment, seems like this should be {{form.event_privacy.1.tag}} and 2nd radio button should be {{form.event_privacy.0.tag}}. Looks like we've changed the variable name from is_public to event_privacy, leading to the mixup.

openCurrents/views.py
```diff
@@ -1258,117 +1258,132 @@ def _create_event(self, location, form_data):
...
-                except Exception as e:
-                    logger.error(
-                        'unable to send transactional email: %s (%s)',
-                        e.message,
-                        type(e)
+        if (coord_user.id != self.userid):
+            # send an invite to coordinator
+            # TODO (@danny):
+            #   - is invite-admin the correct template to use, event for
+            #     existing users?
```
dannypernik: @nickolashe I've created a new template based on the former one, as the message is a bit different for an existing admin. The template is called `change-event-coordinator` and I need one more variable called `EVENT_NAME` passed into the email. I believe I've made the correct changes here:
https://github.com/opencurrents/opencurrents/pull/627/commits/7fd8a0756ac09ed4d0bb20a855f969c63705bd87

openCurrents/forms.py
```diff
@@ -214,12 +214,29 @@ def clean_org_status(self):
...
+        orgid = kwargs.pop('org_id')
+        self.userid = kwargs.pop('user_id')
         self.org = Org.objects.get(id=orgid)
+
+        # this needs to be called first to get access to self.fields
         super(CreateEventForm, self).__init__(*args, **kwargs)
 
+        # build the coordinator choices list dynamically
+        # set (preselect) initially to admin user
+        coordinator_choices = self._get_coordinator_choices()
```
nickolashe: `_get_coordinator_choices` is a function (defined on line 298), so () means call this function without any arguments

openCurrents/forms.py
```diff
@@ -214,12 +214,29 @@ def clean_org_status(self):
...
+            - kwargs contains orgid and is passed from the CreateEventView
+            - builds coordinator choices list dynamically
+            - sets admin user as the initial choice
+        '''
+        orgid = kwargs.pop('org_id')
+        self.userid = kwargs.pop('user_id')
         self.org = Org.objects.get(id=orgid)
+
+        # this needs to be called first to get access to self.fields
         super(CreateEventForm, self).__init__(*args, **kwargs)
```
nickolashe: this calls `__init__` of forms.Form, yes, and this initializes all the basic stuff like access to fields, which is important here

openCurrents/templates/create-event.html
```diff
@@ -1,4 +1,4 @@
-{% load staticfiles %}
+event_date{% load staticfiles %}
```
nickolashe: this is a typo (need to remove event_date)

openCurrents/forms.py
```diff
@@ -228,58 +245,80 @@ def __init__(self, *args, **kwargs):
...
-        })
-    )
-    date_start = forms.CharField(
+    event_date = forms.CharField(
         label='on',
-        widget=forms.TextInput(attrs={
-            'id': 'date_start',
-            'class': 'center'
+        widget=widgets.TextWidget(attrs={
+            'class': 'center',
```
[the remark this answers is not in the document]
nickolashe: yes, we need to give it a try

openCurrents/forms.py
nickolashe: no, i believe its correct. in the html template 0 and 1 just refer to the position within this choices array. so 0 refers to public, and 1 to private.

openCurrents/templates/create-event.html
nickolashe: i am ok with that if that's preferrable

openCurrents/templates/create-event.html
nickolashe: agreed on confusing name. i think event_public is better. see my comment above regarding the order.

openCurrents/templates/edit-event.html
nickolashe: bump

openCurrents/views.py
```diff
@@ -1309,6 +1309,10 @@ def _create_event(self, location, form_data):
                                 'content': self.org.name
                             },
                             {
+                                'name': 'EVENT_NAME',
```
nickolashe: @dannypernik its not the description we want here, but event.project.name

nickolashe: see how Event model points to a Project model which defines the name

nickolashe: since project is also available through `self.project` here, you can also use `self.project.name`

dannypernik: ok, pushed that change

nickolashe: this should work

openCurrents/forms.py
dannypernik: This is confusing - so public is 0 in html and 1 here? Is there a reason to make them inconsistent?

openCurrents/templates/create-event.html
dannypernik: Cool. I moved labels to HTML and also removed the `{% if ... help_text %}` logic, instead adding text where needed.

openCurrents/templates/create-event.html
dannypernik: Removed from edit_event_form.

openCurrents/forms.py
dannypernik: Just now saw your comment below. I'm happy with however you want to handle this variable.

openCurrents/templates/edit-event.html
dannypernik: The span with `input-sizer` class makes it so that the label and the input are on the same line, and the input fills the remaining width on the line. This span needs to wrap the input in order to work.

PR #642, opened 2017-11-08
Add pending balance on biz-admin

dannypernik:
screen shot 2017-11-08 at 2 52 12 am

Are we happy with this design? Here's the alternative:

screen shot 2017-11-08 at 2 49 34 am


nickolashe:
I like the first one better

dannypernik:
Cool, the first design is the one that's available in `balances`. This branch currently throwing an error in profile because coordinator_firstname no longer exists.

PR #751, opened 2017-11-23
public record

danbok:
Implemented desired functionality, unit tests still in development


openCurrents/interfaces/ledger.py
```diff
@@ -110,9 +111,49 @@ def get_balance(self, entity_id, entity_type='user', currency='cur'):
 
         return credit_total - debit_total
 
+    def get_issued_cur_amount(self, entity_id, period):
+        org = self._get_entity(entity_id, 'org')
```
nickolashe: @dabok if the method accepts id of entity as input param, i am not sure there is any need to look up the entity here, since the query filter below uses entitie's id directly

openCurrents/templates/public-record.html
```diff
@@ -2,317 +2,223 @@
...
+  </div>
+</div>
+
+<script type="text/javascript">
+  function makeRequest() {
+    var params = $.param({
+      'type': $('#public-record-filter').val(),
+      'period': $('#time-filter').val()
+    });
+    window.location.href = "{% url 'openCurrents:public-record' %}" + "?" + $('#filter_form').serialize();
```
nickolashe: @dannypernik notice how Danil is looking up the form by id and calling `serialize` to build up the url params

openCurrents/templates/public-record.html
```diff
@@ -2,317 +2,223 @@
...
-    </div>
+<div off-canvas="menu right shift" id="menu">
+  {% include 'partials/menu.html' %}
+</div>
 
-    <div id="public-export-popup" class="modal center small-12 medium-9 large-7 small-centered columns">
-      <div class="row center">
-        <h3 class="popup-title">Export data</h3>
-      </div>
+<div id="public-export-popup" class="modal center small-12 medium-9 large-7 small-centered columns">
```
nickolashe: @danbok @dannypernik I am not seeing this popup anywhere in the template. Am I missing something?

openCurrents/interfaces/ledger.py
danbok: Makes sense, removed

openCurrents/views.py
```diff
@@ -634,9 +635,27 @@ class InventoryView(TemplateView):
...
+    def get_top_list(self, entity_type, period):
+        if entity_type == 'top-org':
+            return OcOrg().get_top_issued_npfs(period)
+        elif entity_type == 'top-vol':
+            return OcUser().get_top_received_users(period)
+        elif entity_type == 'top-biz':
+            return OcOrg().get_top_accepted_bizs(period)
+
+    def get(self, request, *args, **kwargs):
+        entity_type = request.GET.get('record_type', 'top-org')
```
nickolashe: @danbok why not bind the form to data as the first step here (which would also validate the input)?

openCurrents/templates/public-record.html
danbok: It was there before I started editing, I believe it's shown as added line because I adjusted markup (tabs and spaces)

nickolashe:
@danbok i think the functionality is largely implemented, it'd be good to come up with some unit tests now

openCurrents/interfaces/ocuser.py
```diff
@@ -261,6 +262,19 @@ def get_hours_approved(self):
 
         return admin_actions
 
+    def get_top_received_users(self, period):
+        result = list()
+        users = User.objects.all().select_related('userentity')
```
nickolashe: @danbok lets restrict to users with an existing entity here since there can be some special cases where users do not have one (like django's admin user - i was getting an error because of this one)

openCurrents/templates/public-record.html
nickolashe: yeah i think @dannypernik will be able to clarify for us

danbok:
Finished with changes and unittests

openCurrents/templates/public-record.html
dannypernik: @nickolashe at first I designed public-record with a button to export data. I've removed the button but the popup still exists. I believe we'll want this feature eventually

openCurrents/templates/public-record.html
dannypernik: I see. Also interesting to see that the filters use a form to update data.

Issue #601, opened 2017-10-27
Home: Amend messaging on sign-up for clarity and distinction between users, orgs and individuals. 


MaxTattenbach:

dannypernik:
@MaxTattenbach check out https://opencdesign.herokuapp.com and scroll down to the form at the bottom of the page. See if the buttons there do the trick in your opinion.

MaxTattenbach:
Unable to access opencdesign.herokuapp.com.   I am getting an error.

Max Tattenbach
Co-creator
www.opencurrents.com

dannypernik:
@MaxTattenbach it's working for me, not sure the problem

dannypernik:
Ready in home_signup branch (PR #759)

dannypernik:
Closing PR in favor of links directly to forms on various landing pages

PR #759, opened 2017-11-27
make 3 signup popups available to home, volunteer, nonprofit, our-story, and...

dannypernik:
…business via include

Related to issue #601 

@nickolashe can you confirm that it's fine to include the 3 popup forms for registering as an individual, nonprofit, or business in all 5 of these landing pages? I've put the popups in a separate file and used `include` to bring them into the templates.


dannypernik:
Closing PR in favor of links directly to forms on various landing pages

PR #766, opened 2017-12-01
Kostya profile additions

nickolashe:


openCurrents/tests/test_npf_admin_profile.py
```diff
@@ -0,0 +1,242 @@
...
+        tests the content of an NPF org profile page
+        """
+        response = self.client.get('/org-admin/')
+
+        # check if users sees NFL org profile page
+        self.assertEqual(response.status_code, 200)
+        #self.assertTemplateUsed(response, 'org-admin.html')
+
+        # assert org/my hours tracked are displayed
+        self.assertContains(response, '<h6 class="one-margin-top no-margin-bottom">Org hours tracked: 4.0</h6>')
```
nickolashe: can probably remove asserts on html formatting since this may change going forward

nickolashe:
@tevtonez looks good

openCurrents/views.py
```diff
@@ -614,6 +614,29 @@ class HoursApprovedView(LoginRequiredMixin, SessionContextView, TemplateView):
...
 
 
+class HoursDetailView(LoginRequiredMixin, SessionContextView, ListView):
+    template_name = 'hours-detail.html'
+    model = AdminActionUserTime
+
+    def get_queryset(self):
+        queryset = None
+
+        if self.request.GET.get('is_admin'):
```
nickolashe: looks good, lets use forms if possible

openCurrents/views.py
```diff
@@ -1180,6 +1203,14 @@ def get_context_data(self, **kwargs):
         except KeyError:
             pass
 
+        # getting all admins for organization
+        context['org_admins'] = org_admins = []
+        try:
+            org_admins = [u for u in OrgUser.objects.filter(org = self.org) if OcAuth(u.id).is_admin_org]
```
nickolashe: should be `.is_admin_org()`

openCurrents/views.py
```diff
@@ -1180,6 +1203,14 @@ def get_context_data(self, **kwargs):
         except KeyError:
             pass
 
+        # getting all admins for organization
+        context['org_admins'] = org_admins = []
```
nickolashe: what is the result of this line?

openCurrents/views.py
```diff
@@ -1195,36 +1226,50 @@ def get_context_data(self, **kwargs):
...
-        ])
+        # calculating pending hours for every NPF admin
+        context['hours_pending_by_admin'] = []
+
+        for admin in org_admins:
+            pending_by_admin = 0
+            hours_pending = {admin.user.id : pending_by_admin }
+
+            for x in OrgAdmin(admin.user.id).get_hours_requested():
+                pending_hours = (x.usertimelog.datetime_end - x.usertimelog.datetime_start).total_seconds() / 3600
```
nickolashe: i think we have `diffInHours` function we can use here

nickolashe: one way to write this would be to use `reduce` on `OrgAdmin(admin.user.id).get_hours_requested()` while summing up `diffInHours`
http://book.pythontips.com/en/latest/map_filter.html#reduce

openCurrents/views.py
```diff
@@ -1195,36 +1226,50 @@ def get_context_data(self, **kwargs):
...
+                if admin.user.id == self.user.id:
+                    time_issued_by_logged_admin += event_hours
+
+                # adding to the dictionary with admin's approved hours
+                amount_issued_by_admin[admin.user.id] += event_hours
+
+
+            context['issued_by_logged_admin'] = round(time_issued_by_logged_admin,2)
+            amount_issued_by_admin[admin.user.id] = round(amount_issued_by_admin[admin.user.id], 2)
+            context['issued_by_admin'].append(amount_issued_by_admin)
```
nickolashe: can we sort this list by number of hours issued in descending order? so the buttons start with the admin that issued most

openCurrents/views.py
```diff
@@ -1180,6 +1203,14 @@ def get_context_data(self, **kwargs):
         except KeyError:
             pass
 
+        # getting all admins for organization
+        context['org_admins'] = org_admins = []
+        try:
+            org_admins = [u for u in OrgUser.objects.filter(org = self.org) if OcAuth(u.id).is_admin_org]
+        except:
```
nickolashe: can we be explicit about which exception we are aiming to catch here

openCurrents/views.py
nickolashe: does the assignment take place from right to left?

openCurrents/views.py
```diff
@@ -1195,36 +1226,50 @@ def get_context_data(self, **kwargs):
...
-        org_event_user = dict([
-            (event.id, set())
-            for event in Event.objects.filter(project__org__id=self.org.id)
-        ])
+        # calculating pending hours for every NPF admin
+        context['hours_pending_by_admin'] = []
+
+        for admin in org_admins:
+            pending_by_admin = 0
+            hours_pending = {admin.user.id : pending_by_admin }
```
nickolashe: lets only include admins that tracked any hours - no need to include those that did not

openCurrents/views.py
```diff
@@ -1195,36 +1226,50 @@ def get_context_data(self, **kwargs):
...
+            context['hours_pending_by_admin'].append(hours_pending)
+
+
+        # calculating approved hours for every NPF admin and total NPF Org hours tracked
+        context['issued_by_admin'] = []
+        context['issued_by_logged_admin'] = time_issued_by_logged_admin = issued_by_all = 0
 
-        issued_by_all = 0
-        issued_by_admin = 0
+        for admin in org_admins:
```
nickolashe: similar set of comments applies here as for pending hours above

openCurrents/views.py
```diff
@@ -1195,36 +1226,50 @@ def get_context_data(self, **kwargs):
...
+                # adding to current admin's approved hours
+                if admin.user.id == self.user.id:
+                    time_issued_by_logged_admin += event_hours
+
+                # adding to the dictionary with admin's approved hours
+                amount_issued_by_admin[admin.user.id] += event_hours
+
+
+            context['issued_by_logged_admin'] = round(time_issued_by_logged_admin,2)
+            amount_issued_by_admin[admin.user.id] = round(amount_issued_by_admin[admin.user.id], 2)
```
nickolashe: seems like this data is duplicating `time_issued_by_logged_admin`

openCurrents/tests/test_npf_admin_profile.py
tevtonez: this one is fixed already.

openCurrents/views.py
tevtonez: this line just assigns an empty list to both: context['org_admins'] and org_admins
then I'd work with org_admins list iterating through it, but I think we can just use  context['org_admins'] only.
I'll update the code with my next push

openCurrents/views.py
tevtonez: I've fixed this issue in commit 2a4108c

openCurrents/views.py
tevtonez: @nickolashe it seems that this piece of code fails silently, if it receives wrong input, let's discuss it tonight.

tevtonez: I'll need your help with this one @nickolashe 
lets discuss it during one of our conversations. Thanks!

openCurrents/views.py
tevtonez: implemented with e4e2beddde573928fb1681cfc90415f82fd1d62f

openCurrents/views.py
tevtonez: implemented with e4e2beddde573928fb1681cfc90415f82fd1d62f

openCurrents/templates/hours-detail.html
```diff
@@ -35,64 +36,30 @@ <h4 class="title">
...
-            </div>
-
-            <div class="right small-2 columns">
-              <p class="half-margin-bottom">2.25</p>
-            </div>
-          </div>
-        </div>
-      </div>
-
+{% if object_list %}
```
nickolashe: @tevtonez lets change the name to something more meaningful through view's `context_object_name`

openCurrents/templates/hours-detail.html
```diff
@@ -35,64 +36,30 @@ <h4 class="title">
...
+    {% for object in object_list  %}
```
nickolashe: can use `for ... empty` clause and get rid of the outer `if`

openCurrents/views.py
```diff
@@ -1234,12 +1236,16 @@ def get_context_data(self, **kwargs):
             pending_by_admin = 0
             hours_pending = {admin.user.id : pending_by_admin }
 
-            for x in OrgAdmin(admin.user.id).get_hours_requested():
-                pending_hours = (x.usertimelog.datetime_end - x.usertimelog.datetime_start).total_seconds() / 3600
-                hours_pending[admin.user.id] += pending_hours
+            try:
+                hours_pending[admin.user.id] = reduce(lambda x,y : x + y, [diffInHours(x.usertimelog.datetime_start, x.usertimelog.datetime_end) for x in OrgAdmin(admin.user.id).get_hours_requested()])
+            except TypeError:
+                print "No hours approved for this admin!"
```
nickolashe: lets use logger, i.e:
> logger.debug("No hours approved for admin %s", admin.user.username)

openCurrents/views.py
```diff
@@ -1250,23 +1256,26 @@ def get_context_data(self, **kwargs):
             issued_by_admin = 0
             amount_issued_by_admin = {admin.user.id : issued_by_admin }
 
-            for x in OrgAdmin(admin.user.id).get_hours_approved():
-                event_hours = (x.usertimelog.datetime_end - x.usertimelog.datetime_start).total_seconds() / 3600
+            try:
+                amount_issued_by_admin[admin.user.id] = reduce(lambda x,y : x + y, [diffInHours(x.usertimelog.datetime_start, x.usertimelog.datetime_end) for x in OrgAdmin(admin.user.id).get_hours_approved()])
+            except TypeError:
+                print "No hours approved for this admin!"
```
nickolashe: lets use `logger` anywhere we need to print something (see above)

openCurrents/views.py
```diff
@@ -1250,23 +1256,26 @@ def get_context_data(self, **kwargs):
...
+            if admin.user.id == self.user.id:
+                time_issued_by_logged_admin += amount_issued_by_admin[admin.user.id]
 
-                # adding to the dictionary with admin's approved hours
-                amount_issued_by_admin[admin.user.id] += event_hours
+            if amount_issued_by_admin[admin.user.id] > 0:
+                context['issued_by_admin'].append(amount_issued_by_admin)
 
+                # sorting the list of admins by # of issued hours descending
+                context['issued_by_admin'] = sorted(context['issued_by_admin'], key=lambda d: d.keys()[0])
```
nickolashe: i think we want to sort by values and not keys, since values hold the hours?

nickolashe: > sorted(context['issued_by_admin'], key=lambda d: d.values()[0], reverse=True)

openCurrents/views.py
```diff
@@ -1195,6 +1196,19 @@ def get_context_data(self, **kwargs):
 class OrgAdminView(OrgAdminPermissionMixin, OrgSessionContextView, TemplateView):
     template_name = 'org-admin.html'
 
+    def _sorting_hours(self, list_of_dicts, user_id):
```
nickolashe: @tevtonez can you provide a doc string for what this function does pls

openCurrents/views.py
```diff
@@ -1209,8 +1223,8 @@ def get_context_data(self, **kwargs):
         context['org_admins'] = []
         try:
             context['org_admins'] = [u for u in OrgUser.objects.filter(org = self.org.id) if OcAuth(u.user.id).is_admin_org()]
-        except:
-            pass
+        except (UnboundLocalError, InvalidUserException) as e:
+            print e
```
nickolashe: logger

openCurrents/templates/org-admin.html
```diff
@@ -172,6 +173,34 @@ <h6 class="half-margin-bottom">Upcoming events</h6>
     </div>
   </div>
 
+<br>
+  <div class="left row one-margin-top full-width collapse">
```
nickolashe: @tevtonez this block looks very similar to the following, lets put a TODO to use macros here

openCurrents/templates/profile.html
```diff
@@ -263,6 +285,7 @@ <h6 class="popup-title">
       <h3 class="popup-title">No dollars available</h3>
     </div>
 
+=======
```
nickolashe: looks like a remnant of a previous merge?

openCurrents/tests/test_npf_admin_profile.py
```diff
@@ -0,0 +1,356 @@
...
+import pytz
+import uuid
+import random
+import string
+import re
+
+from unittest import skip
+
+
+class NpfAdminView(TestCase):
```
nickolashe: @tevtonez is there a test that tests for the ordering of buttons?

nickolashe: i am still seeing them ordered in the wrong order

openCurrents/views.py
tevtonez: sure, it's updated

openCurrents/views.py
tevtonez: true, fixed

openCurrents/templates/org-admin.html
tevtonez: done as per our discussion

openCurrents/templates/profile.html
tevtonez: yup, I saw it and wanted to fix during the work on profile page, but I updated the code already.

openCurrents/tests/test_npf_admin_profile.py
tevtonez: the order is fixed now:
logged in admin will be shown first, then all the other admins will be show sorted by hours in descending order

openCurrents/interfaces/ocuser.py
```diff
@@ -306,17 +306,34 @@ def get_top_received_users(self, period, quantity=10):
         result.sort(key=lambda user_dict: user_dict['total'], reverse=True)
         return result[:quantity]
 
-    def _get_usertimelogs(self, verified=False):
+    def _get_usertimelogs(self, verified=False, **kwargs):
         # determine whether there are any unverified timelogs for admin
-        usertimelogs = UserTimeLog.objects.filter(
+
+        if 'org_id' in kwargs:
```
nickolashe: i think you can just filter the queryset additionally if the param is present (no need to duplicate the main query in both cases)

openCurrents/tests/test_user_profile.py
```diff
@@ -0,0 +1,300 @@
...
+        datetime_start_2 = past_date + timedelta(hours=3)
+        datetime_end_2 = past_date + timedelta(hours=5)
+
+        # setting 2 approved events for different NPF orgs in the past
+        # _setup_volunteer_hours(volunteer_1, npf_admin_1, org1, project_1, datetime_start_1, datetime_end_1, is_verified = True, action_type = 'app')
+
+        # _setup_volunteer_hours(volunteer_1, npf_admin_2, org2, project_2, datetime_start_2, datetime_end_2, is_verified = True, action_type = 'app')
+
+        # creating ledger entries for CURRENTS
+        ledger_currents_approved_1 = Ledger(
```
nickolashe: @tevtonez lets user `interfaces/ledger.py` to do all transactions, including issuing currents

openCurrents/tests/test_user_profile.py
```diff
@@ -0,0 +1,300 @@
...
+        # @@ TODO @@
+        # THIS TEST IS CRUSHING
+        # self.assertIn('Currents issued: 20.00', response.content)
+        # self.assertEqual(response.context['currents_amount_total'], 20)
+
+        self.assertIn('Active volunteers: 4', response.content)
+        self.assertEqual(response.context['active_volunteers_total'], 4)
+
+        # @@ TODO @@
+        # NEED MORE INFO ON OBJECTS THAT NEED TO BE SET UP TO TEST THIS PART
```
nickolashe: to test this one, you can just issue currents to more volunteers

openCurrents/views.py
```diff
@@ -1173,19 +1203,77 @@ def get_context_data(self, **kwargs):
...
+
+
         # user timezone
         #context['timezone'] = self.request.user.account.timezone
         context['timezone'] = 'America/Chicago'
 
+
+        # getting issued currents
+        try:
+            context['currents_amount_total'] = reduce(lambda x,y : x + y, [x['total'] for x in OcOrg().get_top_issued_npfs(period='all-time') if x['total']>0])
```
nickolashe: if done this way, i think we can simply sum the array instead of using `reduce`.

openCurrents/views.py
```diff
@@ -1173,19 +1203,77 @@ def get_context_data(self, **kwargs):
...
+        except:
+            context['currents_amount_total'] = []
+
+        # getting active volunteers
+        try:
+            context['active_volunteers_total'] = len([x for x in OcUser().get_top_received_users(period='all-time')])
```
nickolashe: why not use `len` directly on the output of `get_top_received_users`?

openCurrents/views.py
```diff
@@ -1173,19 +1203,77 @@ def get_context_data(self, **kwargs):
...
+        except:
+            context['active_volunteers_total'] = []
+
+        # getting currents accepted
+        try:
+            context['currents_accepted'] = reduce(lambda x,y : x + y, [x['total'] for x in OcOrg().get_top_accepted_bizs(period='all-time') if x['total']>0])
```
nickolashe: same as above

openCurrents/views.py
```diff
@@ -1196,6 +1284,14 @@ def get_context_data(self, **kwargs):
         except KeyError:
             pass
 
+        # getting all admins for organization
+        context['org_admins'] = []
+        try:
+            context['org_admins'] = [u for u in OrgUser.objects.filter(org = self.org.id) if OcAuth(u.user.id).is_admin_org()]
```
nickolashe: lets add this as a method to orgs.py

openCurrents/views.py
```diff
@@ -1211,36 +1307,62 @@ def get_context_data(self, **kwargs):
...
+        for admin in context['org_admins']:
+            pending_by_admin = 0
+            form = None
+            hours_pending = {admin.user.id : pending_by_admin }
+            admin_forms = {admin.user.id : form }
 
-        issued_by_all = 0
-        issued_by_admin = 0
+            try:
+                hours_pending[admin.user.id] = reduce(lambda x,y : x + y, [diffInHours(x.usertimelog.datetime_start, x.usertimelog.datetime_end) for x in OrgAdmin(admin.user.id).get_hours_requested()])
```
nickolashe: lets add this a method get_hours_requested_total to OrgAdmin

nickolashe: also change to using `sum`

openCurrents/views.py
nickolashe: calling it get_org_admins()

openCurrents/views.py
nickolashe: lets factor this out as a method in OcOrg called `get_currents_accepted_total`

nickolashe: actually lets put this (and other stats such as this one) into a separate module called community

openCurrents/views.py
nickolashe: lets put this in new module community

openCurrents/views.py
```diff
@@ -1211,36 +1307,62 @@ def get_context_data(self, **kwargs):
...
+        # calculating approved hours for every NPF admin and total NPF Org hours tracked
+        context['issued_by_admin'] = []
+        context['issued_by_logged_admin'] = context['issued_by_all'] = time_issued_by_logged_admin = 0
+
+        for admin in context['org_admins']:
+            issued_by_admin = 0
+            amount_issued_by_admin = {admin.user.id : issued_by_admin }
+
+            try:
+                amount_issued_by_admin[admin.user.id] = reduce(lambda x,y : x + y, [diffInHours(x.usertimelog.datetime_start, x.usertimelog.datetime_end) for x in OrgAdmin(admin.user.id).get_hours_approved()])
```
nickolashe: can use sum

openCurrents/views.py
nickolashe: lets factor out to the new module community and call it `get_active_volunteers_total`

openCurrents/views.py
nickolashe: lets factor out to the new module community and call it get_amount_currents_total

nickolashe:
@tevtonez merging this PR for other to test, please a new one (you can use the same branch)

openCurrents/interfaces/ocuser.py
tevtonez: true, amending it...

openCurrents/views.py
tevtonez: @nickolashe I implement this part before I know how a proper query in this case. I assumed that there might be a case when there is no users in the system in zero state and tried to avoid it. Then, after our discussion, I changed the code but left `try:... except:...`. 
I'll remove it after factoring out it to the new module.

Issue #782, opened 2017-12-05
Biz Details form to be added. 


MaxTattenbach:
As agreed we will release with once the businesses can add some details as contact information, address, etc etc.

nickolashe:
biz detail form is in develop now

nickolashe:
moved to "To be tested"

nickolashe:
@dannypernik @MaxTattenbach has this been tested? please close if so.

dannypernik:
@nickolashe 3 fields display "None" as the default value instead of blank
<img width="745" alt="screen shot 2017-12-12 at 6 21 29 pm" src="https://user-images.githubusercontent.com/3347095/33915526-6eb52840-df69-11e7-8489-5c24690f6a74.png">

dannypernik:
@nickolashe "Introduce your business" field does not save

dannypernik:
Does not check if website is valid - not required for MVP

nickolashe:
are you sure intro is saved? i think it was on my end

MaxTattenbach:
@dannypernik @nickolashe added a dependency to this issue, to complete this issues form must be displayed in the marketplace so users can access biz information.

dannypernik:
@nickolashe I've designed a dropdown to display biz details on `biz_details` form (#837). See that PR for the variables we need. If you let me know the variable names, I can add them to the HTML.

nickolashe:
please note the following discrepancies here:
1. biz details on marketplace should not be referred to as a form - its not taking any user input
2. this issue (#782) is not blocked by anything, #837 does not depend on it

MaxTattenbach:
@nickolashe understood on dependencies. Looks good.

PR #793, opened 2017-12-06
Biz details

nickolashe:


openCurrents/forms.py
```diff
@@ -586,6 +586,48 @@ class EventCheckinForm(forms.Form):
     checkin = VolunteerCheckinField()
 
 
+class BizDetailsForm(forms.Form):
+    def __init__(self, *args, **kwargs):
+        orgid = kwargs.pop('orgid')
```
nickolashe: @dannypernik what do you need the 'orgid' here for?

openCurrents/views.py
```diff
@@ -281,6 +282,7 @@ class AssignAdminsView(TemplateView):
 
 class BizAdminView(BizAdminPermissionMixin, BizSessionContextView, TemplateView):
```
nickolashe: which type of view are we using for Create/EditEvent views? i thought that example was largely comprehensive.

openCurrents/templates/popups/biz-details-popup.html
```diff
@@ -0,0 +1,42 @@
...
+      {% block title %}
+        Add business details
+      {% endblock title %}
+    </h3>
+
+    <p>Let customers know how to reach you:</p>
+  </div>
+{% endblock heading %}
+
+{% block content %}
```
nickolashe: you have not declared the form in html yet

openCurrents/forms.py
dannypernik: Thought it may be needed to associate the data to the org when saving. But is this only to display data from the org?

openCurrents/views.py
dannypernik: CreateEventView(OrgAdminPermissionMixin, SessionContextView, FormView)
and EditEventView(CreateEventView)

dannypernik: Note that I haven't changed the views for BizAdminView

openCurrents/templates/popups/biz-details-popup.html
dannypernik: ah thanks. done now

openCurrents/forms.py
nickolashe: we are not saving any data here. this happens in the view. the only reason it may be needed here is if you are doing some validation specific to the org.

openCurrents/views.py
nickolashe: yes i did note this, this is the reason you are not able to see the form.

nickolashe: in order to be used in the template, the form must be passed in to it (usually called `form` context var). FormView is the only one that does this for you, other ones you have to pass in explicitly (which is not the case here).

openCurrents/templates/popups/biz-details-popup.html
nickolashe: i think you did not push your commit (not seeing it)

openCurrents/templates/popups/biz-details-popup.html
```diff
@@ -13,23 +13,27 @@ <h3 class="one-margin-top half-margin-bottom">
...
+      </div>
     </div>
 
-    <div class="columns">
-      {{ forms.biz_address }}
-    </div>
-  </div>
-
-  {{ forms.biz_intro }}
+    {{ forms.biz_intro }}
```
nickolashe: @dannypernik buttons must be inside the form

openCurrents/views.py
```diff
@@ -280,7 +280,7 @@ class AssignAdminsView(TemplateView):
     template_name = 'assign-admins.html'
 
 
-class BizAdminView(BizAdminPermissionMixin, BizSessionContextView, TemplateView):
+class BizAdminView(BizAdminPermissionMixin, BizSessionContextView, TemplateView, FormView):
```
nickolashe: it should be either TemplateView *or* FormView

openCurrents/templates/popups/biz-details-popup.html
```diff
@@ -0,0 +1,46 @@
...
+
+    <p>Let customers know how to reach you:</p>
+  </div>
+{% endblock heading %}
+
+{% block content %}
+  <form class="" method="post">
+  {% csrf_token %}
+  
+    {{ forms.biz_website }}
```
nickolashe: its `form` that is the name of the context variable, not forms.

nickolashe:
@dannypernik ready for you take a pass:
1. status_msg not showing on biz-admin
2. figure out a way to handle invalid submissions from popup (in terms of bringing user's attention)

dannypernik:
@nickolashe for 2) above we agreed to show form errors on the template. See this commit for my attempt to show either status message or form errors on template:
https://github.com/opencurrents/opencurrents/pull/793/commits/1526f7d95d14921572406ba84380c838b2713c77

With regards to 1) is it not the same issue that we had with status not showing on marketplace? Have you tried the same fix with biz-admin?

nickolashe:
2) looks good
1) no, its not the same. its a formview so it should be showing status messages like other form views we have. there is something else going on i hoped you help figure out.

dannypernik:
Sorry, no idea. I see that error messages are displaying but not status message.  I'm not clear on the purpose of `{% autoescape off %}` but assume it's not the problem.

Also, I changed the error display to match what you had inside the popup (previously I had copied it from create-event.html.) Main difference I'm seeing is that field name is now displayed ("phone: Invalid phone number.") Is every error message field-dependent such that it's redundant to display the field name first?

https://github.com/opencurrents/opencurrents/pull/793/commits/28be075b354da87e179ef32c5ae9b65bf4d66f6f

Finally, I added `required=False` to website field.

nickolashe:
autoescape off is for showing http links in status message properly like on marketplace

nickolashe:
#782

PR #802, opened 2017-12-08
Kostya time tracker update

tevtonez:


openCurrents/views.py
```diff
@@ -935,14 +926,27 @@ def track_hours(self, form_data):
...
-            admin_name = self.request.POST['admin_fname']
-            admin_email = self.request.POST['admin_email']
+
+            admin_name = form_data['new_admin_name']
+            admin_email = form_data['new_admin_email']
+
+            if form_data['new_org'] != '':
+                org = form_data['new_org']
+
             if admin_email:
```
nickolashe: @tevtonez i think in both cases (new admin or new org) we also want to send an email bizdev with the hours so we can keep for record

nickolashe: or may be this is already happening?

PR #810, opened 2017-12-09
Kostya profile additions

tevtonez:


openCurrents/tests/test_approve_hours.py
```diff
@@ -100,23 +96,217 @@ def test_logged_hours_accept(self):
...
+        org_admin_response = self.client.get('/org-admin/')
+        self.assertEqual(org_admin_response.status_code, 200)
+
+        # checking pending hours before approving
+        self.assertListEqual(org_admin_response.context['hours_pending_by_admin'], [{self.npf_admin_1.id: 5.0}])
+
+        # checking total approved hours
+        self.assertEqual(org_admin_response.context['issued_by_all'], 0)
+
+        # approving hours
```
nickolashe: @tevtonez i think its important to assert the creation of the corresponding usertimelog and adminaction records

openCurrents/tests/test_approve_hours.py
```diff
@@ -100,23 +96,217 @@ def test_logged_hours_accept(self):
...
+        self.assertEqual(org_admin_response.context['issued_by_all'], 0)
+
+        self.response = self.client.post('/approve-hours/', {
+                'post-data': self.volunteer_1.username + ':0:' + self.monday +',' + self.volunteer_2.username + ':0:' + self.monday
+                })
+
+        # return to org-amdin after declining
+        self.assertRedirects(self.response, '/org-admin/0/2/', status_code=302)
+
+        # checking approved hours after declining
```
nickolashe: same comment as above applies

nickolashe:
can we assert on the creation of proper records in db. looks good otherwise.

openCurrents/tests/test_approve_hours.py
tevtonez: @nickolashe both handled in 357bd80

PR #822, opened 2017-12-10
Kostya time tracker update

tevtonez:


openCurrents/views.py
```diff
@@ -3314,7 +3320,16 @@ def sendContactEmail(template_name, template_content, merge_vars, admin_email, u
...
-def sendTransactionalEmail(template_name, template_content, merge_vars, recipient_email):
+def sendTransactionalEmail(template_name, template_content, merge_vars, recipient_email, **kwargs):
+
+    # adding session marker for testing purpose
+    #kwargs['session']['transactional'] = kwargs['marker']
+
+    # adding launch function marker to session for testing purpose
+    sess = kwargs['session']
+    marker = kwargs['marker']
+    sess['transactional'] = kwargs['marker']
```
nickolashe: @tevtonez can we opt out of making the call to mandrill in case the test var is present?

tevtonez: @nickolashe good idea, I implemented it in this commit: 
d8f19a7

PR #825, opened 2017-12-11
Sitemap

dannypernik:
@nickolashe I tried implementing Django's sitemap framework but didn't quite get it. Instead I'm making a sitemap available as an xml template. If we can merge this to develop then to master, we'll be able to see if it works for Google search (they require a URL instead of a file upload).

Django documentation: https://docs.djangoproject.com/en/2.0/ref/contrib/sitemaps/
Here's the video tutorial I tried: https://godjango.com/23-robots-and-sitemaps/


nickolashe:
@dannypernik what did not work for you specifically?

Issue #666, opened 2017-11-14
time-tracker: new functionality and unit tests


nickolashe:

MaxTattenbach:
@nickolashe @tevtonez 

Wondering about the functionality of time tracker in these two use cases: 
Volunteer requests Currents with non-registered NPF
Volunteer requests Currents with non-registered admin of a registered NPF

When looking at these issues above I don't see the implementation issues regarding creating a new org user in the database when the form is submitted.  Additionally, a volunteer should get pending Currents and hours should be tracked when the form is submitted. 

I have posted below the link to oC MVP Use Cases, you can find them in Time Tracker. 

https://docs.google.com/spreadsheets/d/15mnrkl4Rc5dsFX8VAGOeuPPfCQTPHyOWV1SVwkXIqNQ/edit?usp=sharing

tevtonez:
@MaxTattenbach @dannypernik 
I've been using the PDF file with instructions for mentioned functionality and testing. It seems that it contradicts the requirements from gSheet. Regarding requested hours PDF doc says that "Record   not   saved   in   DB   but   passed   into   vars   in   email" in the cases when a volunteer requests Currents with non-registered NPF/NPF-admin. Additionally the PDF doc says nothing about creating a new org user in the database in these cases.
Should I rewrite the tests in this situation?

MaxTattenbach:
@tevtonez - yes that is the funcionality needed.  I was unaware the PDF contradicted the MVP Usecases Sheet.  Sorry about that.  
Lets wait to  here from @nickolashe on how to proceed w/ this one.

nickolashe:
existing org new admin = create new user for admin and new usertimelog
new org = just send emails

nickolashe:
@tevtonez @dannypernik @MaxTattenbach is it correct to say that #748 is the only blocker on this epic?

MaxTattenbach:
@nickolashe I'll move to to be tested bucket.   Although I believe we shall figure out the email situation

dannypernik:
@nickolashe @MaxTattenbach @tevtonez also #637. see 2 comments on PR #861

PR #836, opened 2017-12-13
kostya time tracker

tevtonez:


nickolashe:
@tevtonez what's the status of this one?

tevtonez:
@nickolashe as per our discussion:
I added more assertions to the tests, but tests will need further updates due to implementing a feature from #666 (when a user logs hours for non-existing npf-admin)

openCurrents/tests/test_time_tracker.py
```diff
@@ -154,13 +165,12 @@ def test_vol_hours_new_org_new_adm(self):
 
         # posting form
         response = self.client.post("/time-tracker/", {
-            'description':'test manual tracker description',
-            'date_start':'2017-12-07',
+            'description':'test manual tracker non-existing NPF org and admin',
             'admin':'other-admin',
-            'org':self.org.id,
```
nickolashe: @tevtonez it looks like you are not passing org id. what's the reasoning here?

tevtonez: @nickolashe you're right.
I believe for this test it shouldn't be passed, because we're testing non-existing npf org and admin.
Anyhow, I put it back for now, but lets discuss it when we have a chance.

nickolashe: org id is required in the form as far as i can see, unless you also made a change to make it optional?

PR #837, opened 2017-12-13
show-biz-details

dannypernik:
@nickolashe We'll need to pass in variables for each item in the "edit business details" form and booleans for each so that they only display if present:

screen shot 2017-12-13 at 5 41 26 pm


nickolashe:
@dannypernik done. notice that there was already everything present in the template needed for the implementation.

PR #841, opened 2017-12-16
New home

dannypernik:
@nickolashe here's the primary change in this PR. If you like it, please merge:

screen shot 2017-12-15 at 11 51 31 pm


MaxTattenbach:
@dannypernik Lets discuss this a bit more,  Could we put my proposed version side by side to the one above and compare, let's also have the original one.   I don't think people can relate to "creating a Currency". 

First line: It's like bitcoin for doing good. 
Second line:  The world's only digital Currency accessible to all by volunteering. 
Third Line:  Give time, create Currents
Button:  Join openCurrents 

@nickolashe could you hold off on merging this one..

nickolashe:
@dannypernik @MaxTattenbach i think the comparison to bitcoin is inappropriate since we are not using a blockchain as of now.

MaxTattenbach:
@nickolashe what would you compare it to?  We have gotten this in the street when talking to people about openCurrents. O "it's like bitcoin for volunteers" or similar.  

Note we do not believe our early adopters are technical persons that will argue about blockchain etc. We are aiming to be an alt-coin which bitcoin is. 

Happy to discuss

dannypernik:
@nickolashe @MaxTattenbach how about "It's like getting paid for doing good"

Max I'm happy to change "Let's create a currency". How about "openCurrents is a growing global currency accessible to all - generated by volunteer work."

For the button, what I like about "Give me Currents" is that it's a statement made from the perspective of the website visitor. I think this makes it particularly appealing to click.

MaxTattenbach:
@dannypernik  I suggest we sample with a few folks layouts.   I like it's like bitcoin for doing good. I do think we there is some association with it whether or not on a blockchain. 

Not crazy about getting paid or give me Currents - but happy to use what people like most.

nickolashe:
I am not sure on what to compare to exactly, perhaps "its  like getting rewards or air line miles for doing good".

I understand most of the user base is non-technical, however I making the point that it'll be really hard to back this comparison for those that know what bitcoin is.

nickolashe:
is still in discussion? how about we create a new label for issues that track features still in discussion?

dannypernik:
screen shot 2018-01-02 at 4 19 09 pm

here's the new agreed content.

Issue #497, opened 2017-09-19
Offer Delete form


nickolashe:

dannypernik:
@nickolashe trying to remember what we said for the design of this one. Are we good with having a popup displaying "Are you sure you want to delete?" on biz-admin page? The button inside the popup would be a submit button.

nickolashe:
yes i think this will work here

dannypernik:
Cool, popup is in biz-admin on develop

PR #850, opened 2017-12-17
delete offer

nickolashe:


nickolashe:
@dannypernik what was your plan on getting delete offers to work? all i see is some popup that has no reference to a given offer.

nickolashe:
#497

dannypernik:
@nickolashe my thought was to have the following button in the mini menu for each offer:
```
<a class="delete-offer-popup_open">
     <li class="alert"> Delete offer </li>
</a>
```
then having that button link to a delete offer popup. However, is that not going to work? Is this the same situation as displaying receipt where we would need a popup for every offer?

nickolashe:
in order to delete a given item we need to pass in its id. note that there is a builtin view in django for this purpose, as we discussed:
https://docs.djangoproject.com/en/2.0/ref/class-based-views/generic-editing/#deleteview

dannypernik:
@nickolashe got the offer.id passing in correctly. Getting this error on delete-offer:
screen shot 2017-12-19 at 10 28 24 pm

nickolashe:
@dannypernik:
1. i pushed the rest of the implementation for the delete view. please see https://github.com/opencurrents/opencurrents/pull/850/commits/ea3c85d178cc8e1e4dec90dbef39a017e3e42238 on what remained to be done in order to get it to work.
2. i think its important to decide what happens to the existing redemptions for the deleted offer. lets hold off on merging this in the meantime.

dannypernik:
Ok great thanks, I just took a look. It seems we should keep existing redemptions when deleting an offer. Previous redemptions should not be automatically declined just because a business cancels the offer after the fact. A business could manually decline all pending offers if it so chooses. @MaxTattenbach if you'd care to comment?

nickolashe:
1. agreed on previously approved redemptions
2. i don't think we have an interface for manually declining pending, do we?

MaxTattenbach:
@nickolashe @dannypernik agree previous redemptions should remain valid.  

@nickolashe no there is no option to manually decline pending redemptions but that should be fine

dannypernik:
@nickolashe @MaxTattenbach will code this design if we're happy with it:
review purchases

nickolashe:
@dannypernik LGTM

nickolashe:
still working on making sure existing offers are preserved - this needs extra works because it deletes all dependent objects by default

dannypernik:
@nickolashe do you have an issue open for your last comment above?

nickolashe:
#956

Issue #840, opened 2017-12-15
Market Place:  Let's add links to the details of the businesses. 


MaxTattenbach:
![image](https://user-images.githubusercontent.com/26234440/34064267-7b447344-e1bd-11e7-9ae7-c986a5d0288d.png)

dannypernik:
@nickolashe when saving URLs from biz-details form, do we save in a standard format? For example, what happens if some users input http://, others www, and others none? If http or https isn't present in user input, is it safe to assume http:// ? If website has SSL, am I correct in thinking http will forward to https?

I think it makes sense to check for http/https in URL and add http:// when saving variable if neither is present.

nickolashe:
we can validate url input, but we are not doing that now.
i think the easiest would to put a memo in the form for the user to include http:// and use django built url validator

nickolashe:
we are now with the above commit

MaxTattenbach:
@dannypernik  note what happens after clicking on the website link. 
![image](https://user-images.githubusercontent.com/26234440/34277167-45c114b4-e66a-11e7-9903-769b4f0ab1f3.png)

dannypernik:
@MaxTattenbach this is because the website was saved for Fast Folks before @nickolashe changed the website field to validate URL and add http:// if not present.

Max, I made your m.tattenbach an admin of Fast Folks. Please try changing the website for Fast Folks by simply opening the Edit business details popup, then clicking Save. It should update the URL even if you don't change anything. You can check by reopening edit popup, and http:// should be there. Then go back to marketplace and see if the link is working for you.

We won't face this problem on master since all of these changes are getting merged at the same time.

MaxTattenbach:
@dannypernik thanks! That worked.

PR #855, opened 2017-12-18
biz details

nickolashe:
#840


nickolashe:
@dannypernik what's the status here? url validation you requested was implemented 3 days ago

dannypernik:
@nickolashe thanks, just saw your reply on the corresponding issue. The URLfield seems to be working well. However, the biz-details form is returning "Invalid phone number" if no number is entered. It doesn't save the updated email if this error is returned.

PR #873, opened 2017-12-19
fix #846, update unittests

### LINKED ISSUE #846: Live Dashboard: Hours are not been recorded when users don't have a existing password or active account openCurrents.
@dannypernik @nickolashe;  seems to me that we want to track hours on live dashboard no matter if a person is on the platfrom or not.  

Note despite hours not tracked new users are being created in the DB although hours aren't being associated. 

Image #1 checked everyone in. 

image

Image #2. Went back to org and then back into the live dashboard and all the new users were unselected.     Note I added a name for one of the Dustin's when invited to the event to see if not adding name when adding new user was the problem, but it's not.   

image

Image #3 

Created a new event inviting two existing users in the database.  max@opencurrents.com and Max@opencurrents.com (capital M).   Note that Max@opencurrents.com (capital M) was previously created in the database from the previous event, but has not yet created a password. 

Hours were not tracked for Max@opencurrents.com (capital M) 

image

Image #4
When going back to org profile after checking both in and coming back to the event preview (capital M) is unselected.   No hours were recorded for (Capital M)

image

@nickolashe it seems that we should also solve the Capital letter issue in the database.  Perhaps kostya can take a lead there.

tevtonez:


tevtonez:
@nickolashe I think we can delete this branch

PR #880, opened 2017-12-20
Change 'accepted' to 'redeemed' on profile and public-record

dannypernik:


nickolashe:
@MaxTattenbach @dannypernik i've implemented the total count on profile, which is now equal to pending + accepted. it is taking more time than i realized though, since we also have to adjust the public record page to match the count by listing pending as well there. let me know what you think.

dannypernik:
@nickolashe seems worthwhile to add pending to public-record.

nickolashe:
@maxtattenbach what about you?

nickolashe:
@tevtonez any idea why the unit tests are still failing?

tevtonez:
@nickolashe yes.
it looks like fixes for these tests didn't get to this branch.
For example I've changed test_npf_admin_profile.py by adding assertion on OreredDict and in this branch it's still asserting a regular list of dictionaries. The same happens to other tests.
Also it looks like views are outdated too on this branch.

Should I merge the code from develop to this branch and resolve the conflicts?

MaxTattenbach:
@nickolashe sorry for late reply.   Yeah I think in Public Record we should display pending.

MaxTattenbach:
@nickolashe not sure if this is ready for testing.   Doesn't seem to be displaying Currents redeemed at the moment. 

image

image

nickolashe:
The  public record part is not working yet

PR #893, opened 2017-12-22
Create org from profile

dannypernik:


openCurrents/forms.py
```diff
@@ -206,25 +206,31 @@ def clean(self):
         contact_email = cleaned_data['contact_email']
 
 
-class OrgSignupForm(forms.Form):
```
dannypernik: @nickolashe I repurposed the OrgSignupForm - which had been used for the deprecated org-signup.html. Sound good?

dannypernik:
@nickolashe I had miscategorized this as "PR ready to merge". I've now moved to "PR needs backend"

nickolashe:
@dannypernik @MaxTattenbach is this MVP?

dannypernik:
@nickolashe we thought this should be MVP so that a normal user could create a business and post an offer without starting over with a new account.

nickolashe:
ok, lets include this in the next rollout to master

Issue #50, opened 2017-06-03
Live dashboard: Add new user should check them in


nickolashe:

harmonrky:
New user should automatically be checked in when added @nickolashe

nickolashe:
@harmonrky Good point

nickolashe:
Adding new user is working in general, with the only exception right after registering an existing user

PR #896, opened 2017-12-22
lower case emails; various fixes

nickolashe:


nickolashe:
@dannypernik this introduces fixes; some of which are related to prior overwriting `event_register_live` request. one remaining thing i am having trouble with is clearing status messages (e.g. `vol-exist`) from the screen upon successful subsequent action on live-dashboard. can you assist pls.

dannypernik:
@nickolashe seems like separate issues. Currently I'm unable to add volunteers even if no error has been triggered first

dannypernik:
@nickolashe this should take care of it:
https://github.com/opencurrents/opencurrents/commit/0643a5051863074917b29664c32340b435566aff

Also seems like we never got issue #50 working - automatically check in volunteer when adding new volunteer to live-dashboard during event. Would it be straightforward to take care of this now?

Finally, what's the reasoning behind this message if admin tries to add existing volunteer:
screen shot 2017-12-21 at 10 52 11 pm

nickolashe:
@dannypernik 
1. yes, i'll take a look at automatically checking in
2. that error was meant for something else, specifically if checkin request succeeds but subsequent requests fail. i'll come up with a fix.

nickolashe:
in fact we were not reporting request errors at all, so if there was an error checking someone in it would just turn on the checkbox and fail to report the fact that checkin failed

dannypernik:
good catch

nickolashe:
@dannypernik 1 and 2 done, can you give it a try? merging now so @tevtonez can begin on live-dashboard unit tests.

PR #909, opened 2017-12-27
Confirmation email to volunteer with or without optional message

dannypernik:


nickolashe:
@dannypernik nice work, merging

PR #910, opened 2017-12-27
include pending currents 

nickolashe:
(follow-up to #880)

### DIFF
--- a/openCurrents/interfaces/bizadmin.py
+++ b/openCurrents/interfaces/bizadmin.py
@@ -13,7 +13,7 @@ from openCurrents.models import \
 
 from openCurrents.interfaces import common
 from openCurrents.interfaces import convert
-from openCurrents.interfaces.orgs import OrgUserInfo
+from openCurrents.interfaces.ocuser import OcUser
 from openCurrents.interfaces.ledger import OcLedger
 
 import pytz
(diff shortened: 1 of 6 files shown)


nickolashe:
@tevtonez FYI, as of this change, biz current counts now include pending redemptions in addition to accepted. some tests in public-record and profile are failing and may require corresponding changes.

tevtonez:
@nickolashe it appeared that test was failing because the format of returned currents number has been changed from `20.00` to `20.0`
For now I updated the tests to get green pass with `20.0` format, but if we want to have 2 zeros after the comma, we have to decide how we implement it (eg. `"{0:.2f}".format(...)`)

nickolashe:
@tevtonez 1.
1. this side-effect was unintentional in fact. We want 20.00. Could you take a look at what caused this and change tests back to the original?
2. Do we have any tests for redeemed current totals yet?

tevtonez:
@nickolashe unerstood
1. I'll do it and will let you know
2. yes, a basic one in scope of profile page: https://github.com/opencurrents/opencurrents/blob/develop/openCurrents/tests/test_user_profile.py#L434

tevtonez:
@nickolashe here is what caused the issue with one decimal point: https://github.com/opencurrents/opencurrents/commit/052cba7b44c5244401cee8670e17b645b9f9d8f3#diff-aa3a177b0c84a3bb6f09b30a074afcf7R31

tevtonez:
@nickolashe I need your help regarding https://github.com/opencurrents/opencurrents/commit/fb6462d0e628c942e63664b36e9e974482f3cbeb#diff-36779519c5fba599e0ee846b49ac78e9R478 in fb6462d

Issue #906, opened 2017-12-27
invite-volunteers: Parse all valid emails from bulk-email and single-email fields


nickolashe:
make sure email is received

nickolashe:
@tevtonez can we make sure this is handled properly on the backend? i am seeing the following discrepancies in the old functionality in production:
1. a separate user is created for both names and emails (if those are space separated)
2. for the user created from email <> is not removed from the email address, if provided

MaxTattenbach:
@nickolashe  I have increased the level of priority of the issue.  Note that email addresses between <> are not being received/invited.  This is the main format that Gmail uses and most everyone using the copy paste function will be using this format.

nickolashe:
Got it. This will also mean we’ll need to “replay” this step in production once the fix is pushed.

tevtonez:
@nickolashe I see on master it's implemented like this: https://github.com/opencurrents/opencurrents/blob/master/openCurrents/views.py#L1684
Basically, backend takes anything that user enters into text area, splits it by commas and doesn't distinguish names from emails.

if we're going to implement `username <user@email.com>` thing, we have to cover at least 3 options 
```John Snow <mrsnow@snow.com>
Snow <js@snow.com>
js@snow.com```
so the possible regex would look like 
```(?:([^<>\s]+)(?: )([^<>\s]+)(\s<[^<>]+@[^<>]+>))|([^<>\s]+) <([^<>]+@[^<>]+)>|([^<>]+@[^<>]+)```

please, let me know if we're going to implement it this way, thanks!

nickolashe:
@tevtonez i think the primary use case is when they copy email list from gmail, which means it may be 1 and 3 above + <email> alone

nickolashe:
what about current implementation in develop? i think its doing the same thing? lets implement the proper way in develop considering the above three options i mentioned.

MaxTattenbach:
@nickolashe let me know once this has been merged and where to test.  Will we be testing in develop?

nickolashe:
@MaxTattenbach in develop

MaxTattenbach:
@nickolashe @tevtonez functionality is working properly.   Thanks @tevtonez

MaxTattenbach:
@nickolashe @tevtonez  seems that this was also implemented in the invite volunteers from events which is proper functionality.    Great!

MaxTattenbach:
@nickolashe  not sure what I am supposed to test here?

[...]

PR #913, opened 2017-12-28
Kostya profile additions

tevtonez:
@nickolashe please review float numbers formatting implementation per our yesterday's discussion, thanks!
Email parser is implemented now.

### DIFF
--- a/openCurrents/interfaces/bizadmin.py
+++ b/openCurrents/interfaces/bizadmin.py
@@ -39,7 +39,7 @@ class BizAdmin(object):
             currency=currency
         )
 
-        return round(balance, 3)
+        return balance
 
     def get_balance_pending(self):
         '''
@@ -51,7 +51,7 @@ class BizAdmin(object):
             active_redemption_reqs
         )
 
-        return round(total_redemptions, 3)
+        return total_redemptions
 
     def get_offers_all(self):
         offers = Offer.objects.filter(
--- a/openCurrents/interfaces/community.py
+++ b/openCurrents/interfaces/community.py
@@ -51,7 +51,7 @@ class OcCommunity(object):
     def get_biz_currents_total(self):
         """
         returns total currents accepted in the system
-        note: accepted = approved + redeemed
+        note: accepted = approved + pending
         """
         currents_total = sum([
             x['total']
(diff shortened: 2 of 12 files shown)


nickolashe:
@tevtonez rounding looks good. just to be clear, are you done with email parsing?

tevtonez:
@nickolashe 
> are you done with email parsing?
yes, I still recommend to show "correct" formats of emails that may be used in bulk field

tevtonez:
oops, I accidentally pushed broken test, fixing...

tevtonez:
The work on this branch is finished for now and can be merged...

nickolashe:
@dannypernik what do you think on Kostya's comment above regarding email formats?

dannypernik:
I've added a placeholder with example emails. @tevtonez will the email parser identify emails separated by newlines?

tevtonez:
@dannypernik no, only commas are counted as a separator.

dannypernik:
@tevtonez @nickolashe I believe we want to parse all valid emails from any text copy/pasted into bulk email box. The most pertinent example is when someone copies emails from Gmail, which displays this format: `Danny Pernik <danny@opencurrents.co>, Nicko Shestopalov <nicko@opencurrents.co>` but seems with regex we could just as easily parse emails from any text?

tevtonez:
@dannypernik @nickolashe the request in #906 says that we should extract not only emails but first and last names (which we do currently)
Currently we're splitting string by commas and it's shouldn't be hard to split it by new lines too, but the parser may not work in cases when user accidentally provides bulk emails in format like below (note a new line between Firstname2  and Lastname2):
```
Firstname1 Lastname1 <email1@eeee.ee>, Firstname2 
Lastname2 <email2@eeee.ee>, Firstname3 Lastname3 <email2@eeee.ee>,
Firstname4 Lastname4 <email2@eeee.ee>, 
```
so limiting to only one possible format seems more preferable, since the users will have less options in this case thus the possibility to mix the options and make a mistake is lower.

dannypernik:
@tevtonez @MaxTattenbach @nickolashe I don't believe the intention of #906 was to set user's first and last name, but simply that a user could copy names into the box and we could still parse all emails. So you can ignore all text that is not in email format.

nickolashe:
That’s correct, we don’t need to use the data to set names.

dannypernik:
@tevtonez reopening #906 with a more specific title

Issue #927, opened 2017-12-29
Can't access admin profile when linking email to org on backed end. 


MaxTattenbach:
@nickolashe I have linked a user on the backend created the org user and then granted him access but this particular user can't access the org. 

![image](https://user-images.githubusercontent.com/26234440/34427249-c530f854-ec05-11e7-9ec9-38d86c5c1874.png)

![image](https://user-images.githubusercontent.com/26234440/34427274-25d87fa6-ec06-11e7-9113-b5c3592efd7b.png)

![image](https://user-images.githubusercontent.com/26234440/34427284-4115b310-ec06-11e7-9270-44664305e1fc.png)

![image](https://user-images.githubusercontent.com/26234440/34427255-d71a4afc-ec05-11e7-8424-9c63e94e40d9.png)

dannypernik:
@MaxTattenbach I guess you're working on oC.com here since Costas verdes is org id 61 there. Perhaps you forgot to click "Save" after assigning your user to admin group, as your affiliation is not appearing there:
<img width="879" alt="screen shot 2018-01-02 at 6 56 54 pm" src="https://user-images.githubusercontent.com/3347095/34506015-bf6c2032-efee-11e7-9c2d-95aa27e87fbc.png">

nickolashe:
@tevtonez we need to make sure instead of a runtime error user is directed to `/403` in this case.

tevtonez:
@nickolashe good point! I added the needed code in acdf72a and 53be9e1

nickolashe:
@MaxTattenbach in develop

MaxTattenbach:
@nickolashe @tevtonez this is working now!

PR #958, opened 2018-01-03
#927 - add 403 redirect if admin doesn't belong to a group and tries to reach org-admin page

tevtonez:


nickolashe:
@tevtonez i don't understand why this diff is showing a bunch of commits that are already in develop, but i am guessing that's probably due to rebasing head. looks like this is still ok to merge.

PR #962, opened 2018-01-04
Add facebook pixel with advanced tracking

dannypernik:
@nickolashe I took a best guess as to what variables to add to Facebook Pixel to track our signup form. I used the `name=` field from HTML. Does that seem right? Here are FB's instructions:

screen shot 2018-01-04 at 2 53 11 pm

### DIFF
--- a/openCurrents/templates/check-email.html
+++ b/openCurrents/templates/check-email.html
@@ -1,13 +1,6 @@
 {% extends "layouts/base.html" %}
 {% load staticfiles %}
 
-{% block head %}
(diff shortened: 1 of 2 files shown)


nickolashe:
@dannypernik the variables probably make sense, although we'll have to see it in the works. i am not seeing the full js code from the snapshot, however? did you include it elsewhere?

dannypernik:
@nickolashe I had added the standard pixel code previously. The advanced pixel just adds a few lines. See the lines before and after this PR's change to head.html

PR #971, opened 2018-01-07
add hubspot tracking to landing pages

dannypernik:
@nickolashe I knew you wouldn't like this unDRY code, so I converted all landing page templates to extend a new layout called home-base.html in develop. However, @MaxTattenbach is looking to start tracking via Hubspot ASAP, so I put this code in a PR straight to master so that you wouldnt have to review all the changes to the landing pages in order to merge. In develop branch, this tracking code only appears in home-base.html and base.html. It is not feasible for one of these templates to extend the other, so it's necessary to have the tracking code twice.
 
(This temporary unDRY code saves you from needing to review this: https://github.com/opencurrents/opencurrents/commit/40f471368e54423bca57f99fcf65ff67dc523cc0)


nickolashe:
@dannypernik

1. Will there be a conflict merging develop later?
2. Why does it need to be in two base templates and its unfeasible to have just one?

dannypernik:
@nickolashe 

1. I'll be able to easily resolve any conflicts later.
2. home-base and base.html are very different layouts. It would add confusion to combine them and I don't see value in doing so:

https://github.com/opencurrents/opencurrents/blob/develop/openCurrents/templates/layouts/home-base.html
https://github.com/opencurrents/opencurrents/blob/develop/openCurrents/templates/layouts/base.html

nickolashe:
1. I am not seeing the benefit of doing it one way in master vs a different way in develop. Seems like the review time would be negligible to the time needed to resolve.
2. How about having them both extend some kind of template that initiates the script?

dannypernik:
@nickolashe just tried making a PR from original branch but it has stuff from develop too: https://github.com/opencurrents/opencurrents/pull/972

Issue #977, opened 2018-01-08
Status message “none” displaying when visiting marketplace and no status_msg defined


tevtonez:
<img width="1049" alt="screen shot 2018-01-08 at 10 36 49 am 1" src="https://user-images.githubusercontent.com/343347/34682105-33302334-f49e-11e7-962f-38454382dbbb.png">

Issue #967, opened 2018-01-07
Org-Admin:  Admin of of org is not displaying nothing under Hours Approved until he approves an hour. 


MaxTattenbach:
Until NPF admin approves hours is he able to see hours approved in for his NPF.   He should be able to see Hours Approved regardless if he has accepted hours or not. 

![image](https://user-images.githubusercontent.com/26234440/34645727-90aafcb2-f31a-11e7-96a1-1e89c4c491f2.png)

After "Max Tattenbach"  new registered admin approves hour are the buttons displaying under Hours Approved.
  
![image](https://user-images.githubusercontent.com/26234440/34645789-41572e7c-f31c-11e7-99b5-6ad9e3eac31f.png)

nickolashe:
good find @MaxTattenbach

MaxTattenbach:
@tevtonez issues persist.  Current admin needs is not displaying hours pending approval or hours tracked until he has approved or one hour pending.

tevtonez:
@MaxTattenbach @nickolashe it doesn't seem that my code get to develop, this part shouldn't be in views.py
![screenshot 2018-01-08 23 40 54](https://user-images.githubusercontent.com/343347/34696031-6e628a76-f4cd-11e7-961b-7d96f150dd93.png)

tevtonez:
views.py on my branch: https://github.com/opencurrents/opencurrents/blob/kostya_profile_additions/openCurrents/views.py#L1527

views.py on develop:
https://github.com/opencurrents/opencurrents/blob/develop/openCurrents/views.py#L1511

nickolashe:
@tevtonez No diffs between develop and kostya_profile_additions:

https://github.com/opencurrents/opencurrents/compare/develop...kostya_profile_additions?expand=1

Issue #966, opened 2018-01-07
Org-Admin:  Hours Pending Approval on Org Admin Profile only displaying when current admin has hours Pending.


MaxTattenbach:
All Org admins should always be able see hours pending of other admins.  Currently current Admin most have hours pending to see other admins pending hours. 

Note screenshot below is of two admins accounts, in which one has hours pending and the other doesn't. 

![image](https://user-images.githubusercontent.com/26234440/34645686-b72e4af2-f319-11e7-8c2b-c717caf1708b.png)

Now when both admins have hours pending. 

![image](https://user-images.githubusercontent.com/26234440/34645785-17651ef8-f31c-11e7-8381-08f9d801b685.png)

Don't get names confused - there is a Max Tattenbach which is only a volunteer.

nickolashe:
good find @MaxTattenbach

MaxTattenbach:
@tevtonez  See that "Hours pending approval" are not displaying the hours.    Max Costas Tattenbach submitted hours to admin NS from Casa Marianella.    Admin Max Dot Gmail can't isn't displaying those hours. 

![image](https://user-images.githubusercontent.com/26234440/34695222-a20db8b6-f48f-11e7-9a1a-e4428d65305a.png)

![image](https://user-images.githubusercontent.com/26234440/34695263-ca3f4d36-f48f-11e7-84d0-a22fcb88821e.png)

tevtonez:
@MaxTattenbach I cannot confirm the last issue on develop (please see screenshot)
![screenshot 2018-01-08 23 36 38](https://user-images.githubusercontent.com/343347/34695970-290e2ebc-f4cd-11e7-8fbd-d537bc9097e4.png)

nickolashe:
@MaxTattenbach:
1. what do you mean by the last statement there, "Max Costas submitted hours Max Yahoo from Casa Marienella"?
2. do you think you can change your account names to be something more intuitive as to where it will be clear on the role of the user from the name? (i.e. something along the lines of the fixtures produced by the setup script). reason is i find it difficult to trace the logic of the test cases in some instances.

MaxTattenbach:
@nickolashe @tevtonez functionality is working. 

@nickolashe, yes I will start using more intuitive names.

Issue #949, opened 2017-12-29
Invite Admin: from "Request Currents from another Organization", email should be changed.


MaxTattenbach:
Note that we are using the same email template for new admin when Org is existent or not existent.  We should create a new template for when org is not yet on openCurrents.

Note that Yahoo is not using openCurrents yet.  

Also, note that there is no way we can add pending Currents on the backend if we don't have those variables in the email biz dev receives.  Hours worked, detail of work, etc.   I am not sure if this is even possible to collect as we are currently not storing info on the info on the DB?   

Let's only address the top issue for MVP if the bottom issue/picture is a hard fix. 

![image](https://user-images.githubusercontent.com/26234440/34444459-f222e7e4-ec93-11e7-8a45-79a1a6d950ff.png)

![image](https://user-images.githubusercontent.com/26234440/34444480-1807be8a-ec94-11e7-89fb-26bb7e5bbea7.png)

nickolashe:
@dannypernik could you pick an estimate

MaxTattenbach:
@nickolashe @dannypernik  I am ok for 1/12 although do note that if users record Currents for new organizations then we will be sending false emails stating that the NGO already tracks hours with openC, when they don't.   

@dannypernik if we could create a simple email template for new orgs that would be quick fix.

nickolashe:
@dannypernik

dannypernik:
email template: volunteer-invites-org

<img width="606" alt="screen shot 2018-01-04 at 11 17 24 pm" src="https://user-images.githubusercontent.com/3347095/34596459-73554b34-f1a5-11e7-8705-cc511aad4f3a.png">

nickolashe:
@dannypernik what's left to be done here?

dannypernik:
@nickolashe we'll need to set up logic in views.py to send `volunteer-invites-org` template when "Request Currents from other organization" is clicked and org name and admin email are populated. Will be similar to this but for new orgs:
https://github.com/opencurrents/opencurrents/blob/develop/openCurrents/views.py#L1268

I don't know how we are detecting when Request new org is selected instead of an existing org. @tevtonez did you work on this logic in JS on time-tracker or in views?

nickolashe:
@tevtonez assigning to you (high priority)

tevtonez:
@dannypernik the logic for this is implemented in views

MaxTattenbach:
@tevtonez  Admin is still receiving improper email. 

[...]

PR #981, opened 2018-01-09
restoring fixes for #949 #926, #977, #966, #967

tevtonez:


openCurrents/views.py
```diff
@@ -749,9 +749,11 @@ def get_context_data(self, **kwargs):
         context['user_balance_available'] = user_balance_available
 
         # workaround with status message for ListView
-        if self.kwargs.get('status_msg') is None:
-            context['status_msg'] = ''
+        if 'status_msg' in self.kwargs:
```
nickolashe: @tevtonez this one is not necessary to restore since I moved this logic to SessionContextView

PR #1007, opened 2018-01-16
past events template

dannypernik:


dannypernik:
@tevtonez do you know why this test is failing?

FAIL: test_npf_admins_displayed_in_pending_approved_hours_section (openCurrents.tests.test_npf_admin_profile.NpfAdminView)

tevtonez:
@dannypernik it happens because ampersands in `?is_admin=1&user_id={{admin.id}}&type=approved`  shouldn't be replaced with `&amp;` 
it's fixed on dev branch so I suggest to pull from dev before PR

dannypernik:
@tevtonez if I don't replace `&` with `&amp;` it breaks code recognition in my editor. See screenshot. Is there any way to allow `&amp;`? I believe it's best practice for HTML but open to suggestion.

screen shot 2018-01-17 at 5 11 44 pm

dannypernik:
@tevtonez thanks for looking into it; @nickolashe helped me fix the test to be consistent with `&amp;`. Tests passing!

tevtonez:
@dannypernik no problem! 
+1 for tests passing :)

PR #1016, opened 2018-01-18
Master offer

dannypernik:


openCurrents/templates/marketplace.html
```diff
@@ -9,112 +9,128 @@
...
-                  {{ offer.currents_share }}% of {{ offer.item.name }}
-                </a>
-
-                <span class="grey small-text">
-                  {% if offer.limit != -1 %}
-                    {% if num_remaining > 0 %}
-                      ({{ num_remaining }}
-                    {% else %}
-                      (0
+      {% if offers %}
```
nickolashe: @dannypernik i think you can do the same thing a bit more concisely using `forloop.first` boolean inside the for loop:
https://docs.djangoproject.com/en/2.0/ref/templates/builtins/#for

nickolashe: in other words, i don't think we need the outer if clause in that case

dannypernik: @nickolashe I'm not sure about that, but in considering your comment, I realized I need to make some kind of modification. Consider the case where the master offer is the only offer posted. I'd like the text for `if not offers` to display (https://github.com/opencurrents/opencurrents/pull/1016/files/ba653d230fc344b3a223245639226c3797f212d5#diff-151f6168cb77d4aa0a2ba46277c88353R12). Can we exclude the master offer from `if offers`?

Assuming we excluded master offer in forloop, we'd have to repeat code in forloop and forloop.first. I believe `if offers` is appropriate to display lines 30-35.

openCurrents/templates/redeem-currents.html
```diff
@@ -171,10 +172,18 @@ <h4 class="one-margin-bottom">Confirm exchange</h4>
       </div>
     </div>
 
+    {% if offer.id == master_offer.id %}
```
nickolashe: @dannypernik tried moving the popup inside the form, but still having trouble saving `biz_name`

nickolashe:
@dannypernik Here is a list of tasks we need to do to get this released:
https://github.com/opencurrents/opencurrents/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Master+Offer%22

Issue #1017, opened 2018-01-18
live-dashboard: incorrect behavior when adding to past events


nickolashe:
Invite Volunteers - Event Past.

Volunteers in DB with no password - don't receive event email and don't get added to the event. (Proper functionality should be: added to event to event with no email)
Volunteers in DB with password - don't receive event email and don't get added to the event. (Proper functionality should be: added to event don't receive event email)
New Volunteers not in DB - receive event email "new" and get added to event (Proper functionality should be: added to event don't receive event email)

Issue #1014, opened 2018-01-16
Live Dash Board:  Not Allowing to add volunteer with oC account if box to invite oC is clicked. 


MaxTattenbach:
I am unable to add volunteer via live dash board if he already has an account and the incorrect message dispalys: "Server Error....."  

Note that the status message below the "Server Errror..." is there from the other volunteers that were added, I created a separate issue because that message should be changed. 

When selecting invite box and adding an active volunteer on oC the proper functionality should be:  Add him to the dashboard, don't send invite email and don't display an Error message. 

![image](https://user-images.githubusercontent.com/26234440/35003521-b42ad32a-fab2-11e7-96d2-b734ecba3b3d.png)

nickolashe:
@tevtonez which unit test corresponds to the use case described here?

tevtonez:
@nickolashe I doubt we have a use case for it. oC MVP doesn't contain it either.

nickolashe:
@MaxTattenbach @dannypernik i believe the issue has been fixed. can you give it another try?

PR #1037, opened 2018-01-25
live-dashboard: unit tests

nickolashe:


nickolashe:
@tevtonez are the tests for #992 - #996  and #999 complete at this point?

tevtonez:
@nickolashe I believe it's only #997 that is left to be tested along with parts of the other tests when grey check-in icons turn into blue when users are checked in.

nickolashe:
ok, lets make sure we cover #997 and #1039

nickolashe:
@tevtonez is this PR read to merge once the remaining two paths are complete and passing?

tevtonez:
@nickolashe the test are complete but they will be failing until #1017 is fixed (as per our discussion today)

openCurrents/tests/test_live_dashboard.py
```diff
@@ -0,0 +1,1263 @@
...
+            {
+                'userid':str(self.volunteer_1.id),
+                'checkin':'true',
+            })
+
+        self.assertEqual(post_response.status_code, 201)
+
+        # check there are two checked users
+        response = self.client.get('/live-dashboard/2/')
+        self.assertEqual(len(response.context['checkedin_users']), 2) # THIS ASSERTION FAILS !!!
```
nickolashe: @tevtonez the reason this assert fails is you are asserting checked in users for an event (2) that's different from the one you posted the check in to (3)

openCurrents/tests/test_live_dashboard.py
```diff
@@ -0,0 +1,1263 @@
...
+        # asserting the first volunteer has blue icon
+        processed_content = re.sub(r'\s+', ' ', response.content )
+        self.assertIn('name="vol-checkin-6" value="6" class="hidden checkin-checkbox" checked', processed_content)
+
+        # asserting new user in checkedin_users
+        self.assertIn(new_user_id, response.context['checkedin_users'])
+
+
+    def test_past_event_invite_new_user_invitation_opt_in(self):
+        """
```
nickolashe: @tevtonez having such a summary is really helpful when reviewing. can you please remember to include it for all tests going forward?

openCurrents/tests/test_live_dashboard.py
nickolashe: @tevtonez i think it makes to parametrize the event id for each test to avoid hardcoded bugs such as this one

openCurrents/tests/test_live_dashboard.py
```diff
@@ -0,0 +1,1223 @@
...
+
+        self.client.login(username=self.npf_admin.username, password='password')
+        response = self.client.get('/live-dashboard/2/')
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.context['registered_users']), 3)
+        self._add_user_to_event(self.npf_admin, 2, self.volunteer_4, 3)
+
+        # 1 in the url below means we're mocking emails
+        post_signup = self.client.post('/process_signup/True/True/1/',
```
nickolashe: in POSTing to /process_signup, we need to assert the correct status code and data returned as wel

nickolashe: lets also add a use case same as this one, except user is existing. in such case, invite email should not get sent (see #1014)

openCurrents/tests/test_live_dashboard.py
```diff
@@ -0,0 +1,1223 @@
...
+        self.assertEqual(response.status_code, 200)
+
+        # verify there are correct number of registered users under event
+        self.assertEqual(len(response.context['registered_users']), registered_users_num)
+
+        # assert approved hours from npf admin perspective
+        self.assertEqual(0.0 , self.org_npf_adm.get_total_hours_issued())
+
+        # registering the first volunteer
+        post_response = self.client.post('/event_register_live/{}/'.format(str(event_id)),
```
nickolashe: @tevtonez /event_register_live returns a status code and json data:
https://github.com/opencurrents/opencurrents/blob/develop/openCurrents/views.py#L3109-L3113

lets assert on both here

openCurrents/tests/test_live_dashboard.py
nickolashe: @tevtonez the fore-mentioned use case will be identical for past / current / future events, FYI

openCurrents/tests/test_live_dashboard.py
tevtonez: @nickolashe  it's already there few lines below
```
if expected_user_registration:
            self.assertEqual(post_response.status_code, 201)
            self.assertEqual(len(response.context['registered_users']), registered_users_num + 1)
            self.assertIn('"userid": "{}"'.format(str(volunteer.id)), post_response.content)
            self.assertIn('"eventid": "{}"'.format(str(event_id)), post_response.content)

        else:
            self.assertEqual(post_response.status_code, 201)
            self.assertEqual(len(response.context['registered_users']), registered_users_num)
```

nickolashe: i see. i got the impression its asserting on the payload of /live-dashboard because of where the code is located. all good on this one.

openCurrents/tests/test_live_dashboard.py
```diff
@@ -0,0 +1,1450 @@
...
+        # getting data from live-dasboard
+        response = self.client.get('/live-dashboard/{}/'.format(str(event_id)))
+
+        if expected_user_registration:
+            self.assertEqual(post_response.status_code, 201)
+            self.assertEqual(len(response.context['registered_users']), registered_users_num + 1)
+            self.assertIn('"userid": "{}"'.format(str(volunteer.id)), post_response.content)
+            self.assertIn('"eventid": "{}"'.format(str(event_id)), post_response.content)
+
+        else:
```
nickolashe: why do we need this `else` clause? is not the post state of  /event-register-live that of the user is registered for the event.

tevtonez: @nickolashe it's for assertions in case I'm not expecting a user to be added to 'registered_users'. 
`expected_user_registration` defines if we are expecting user to bee added to event or not

